PR #1, opened 2017-06-28
Documentation suggestions

bendemaree:
Made a commit per change, so you can tweak the group of suggestions as you please!

Notes:
- The motivation page is quite good; consider moving the bullet points onto the home page?
- I don't know what the valid "minimum" priority (largest numeric value) is for an actor's priority, so I left a placeholder in.

I took some liberties. Overall, found the guide quite good, easy to follow, and enjoyable. Have counted so many words. 😁

### DIFF
--- a/docs/source/guide.rst
+++ b/docs/source/guide.rst
@@ -180,6 +180,12 @@ to the main worker process::
 
   $ kill -s HUP 13047
 
+.. note::
+    Subsitute the process ID (PID) of your own main worker process
+    for `13047`. You can find the PID by looking at the log lines
+    from the worker starting up. Make sure to choose the PID from
+    the main process's log line, denoted by ``[dramatiq.MainProcess]``!
+
 The next time your message is run you should see::
 
   Message dropped due to invalid url: 'foo'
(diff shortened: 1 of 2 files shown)


Bogdanp:
Thanks Ben! I took your commits, applied some of my own changes on top and then squished everything into one commit you can see here: https://github.com/Bogdanp/dramatiq/commit/59937c06f58cacf187851a10cfb7c1f9e94fb0fb 

> I don't know what the valid "minimum" priority (largest numeric value) is for an actor's priority, so I left a placeholder in.

The [work queue](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/worker.py#L265) is a `PriorityQueue` so the priority of an actor can be any comparable object. I updated the guide to say the values are limited to the set of positive integers (yay white lies!) :D 

> The motivation page is quite good; consider moving the bullet points onto the home page?

I'll leave it as is for now because I plan to expand that page a little.

Thanks again!

bendemaree:
Right on! Happy to help. :grinning:

slidenerd:
Could you kindly add deployment as a section in the docs where you cover AWS, Docker and Heroku at the least

PR #4, opened 2017-10-30
Provide a URLRabbitmqBroker, configurable via URL

whalesalad:
I'm sure there are style violations or similar here, let me know what I can do to clean this up so it is ready to merge (assuming this addition is desired).

We tend to use ENV variables at FarmLogs to configure our applications. In this sense we often rely on RABBITMQ_URL as a convention.

Fortunately this was easy enough to implement in my own codebase, but I figured others might appreciate this convenience as well. Another nice benefit of this approach is I no longer need to be concerned with `import pika` in my code and can rely wholly on Dramatiq.

This allows you to instantiate a URLRabbitmqBroker with a simple `url` parameter which will get threaded into Pika's [`URLParameters`](http://pika.readthedocs.io/en/0.10.0/examples/using_urlparameters.html)

I also updated the docstring for the regular `RabbitmqBroker` so that it would be a little more clear that the parameters are specifically for [pika connection parameters](http://pika.readthedocs.io/en/0.10.0/modules/parameters.html). (a link might be good too but not sure how your doc rendering system likes to handle that)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -8,3 +8,4 @@ of those changes to CLEARTYPE SRL.
 
 | Username | Name |
 | :------- | :--- |
+| [whalesalad](http://github.com/whalesalad) | Michael Whalen |
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -33,6 +33,9 @@ Brokers
 .. autoclass:: dramatiq.brokers.rabbitmq.RabbitmqBroker
    :members:
    :inherited-members:
+.. autoclass:: dramatiq.brokers.rabbitmq.URLRabbitmqBroker
+   :members:
+   :inherited-members:
 .. autoclass:: dramatiq.brokers.redis.RedisBroker
    :members:
    :inherited-members:
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -22,7 +22,7 @@ class RabbitmqBroker(Broker):
     Parameters:
       middleware(list[Middleware]): The set of middleware that apply
         to this broker.
-      \**parameters(dict): The connection parameters to use to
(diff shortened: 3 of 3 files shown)


Bogdanp:
I like this! Just a couple nitpicks and one process thing:

* Would you mind adding a note to the new class' docstring noting that url params need to include `&heartbeat_interval=0`?  And can you add the new class to `docs/source/reference.rst`?
* Can you rebase on top of origin master and squash your commits into a single commit (preferably with a message like `feature: add URLRabbitmqBroker`)? If you don't know how, I can do this for you when I merge your PR.
* Lastly, can you add yourself to the contributors file in a separate commit? 

Thanks!

whalesalad:
Nice! I'll make some tweaks.

I actually have a question about the heartbeat stuff... right now I am running **with** `?heartbeat=5` because otherwise my connection dies and I get a ton of EOF issues.

Note: my rabbit instances are load balanced behind an Amazon ELB so the heartbeat is needed or the ELB gives up on you.

Bogdanp:
It looks like that may have been a misunderstanding on my part; I thought pika didn't handle heartbeats under `BlockingConnection` but [this](https://github.com/pika/pika/blob/51fd186f1fabe0da8582609d266df1e28e57e3f4/docs/examples/heartbeat_and_blocked_timeouts.rst) makes it seem like it does! So you can skip that. I'll have to remember to remove the recommendation from other places I've added it.

Bogdanp:
I ended up cherry-picking both commits as 75bd1c2a5bbd72ad8115b56f733cb126d93313be and 716aed7d4372a29524bf4506301831e30f0f8a95, respectively. I pushed release `0.10.0` containing these changes and a few others, incl. Ben's fix for the issue you opened.

Thanks again!

whalesalad:
Awesome!

PR #13, opened 2017-11-16
Python35 compatible

jssuzanne:


Bogdanp:
I merged your changes with rebase (so GH isn't going to pick them up). Thanks for your work on this!

I'm going to play around with Python 3.4 support in the `py34` branch. Master, as it exists right now, works with 3.4 but there are a couple failing tests that I need to dig into.

Thanks again!

jssuzanne:
Thanks, I have also start for python 3.4. I am bloking , cause of the same failing tests. I have juste fix the `yield from`  in **setup.py**

Bogdanp:
I didn't have to change the yield from locally. Everything seemed to work fine except the failing tests.

jssuzanne:
Ok,

PR #21, opened 2017-12-25
feature: add support for callbacks

Bogdanp:


dramatiq/middleware/callbacks.py
```diff
@@ -0,0 +1,30 @@
...
+            actor_name = message.options.get("on_success")
+            if actor_name:
+                actor = broker.get_actor(actor_name)
+                actor.send(message._asdict(), result)
+
+        else:
+            actor_name = message.options.get("on_failure")
+            if actor_name:
+                actor = broker.get_actor(actor_name)
+                actor.send(message._asdict(), {
```
Bogdanp: Messages should probably expose `_asdict` as a "public" method.

dramatiq/actor.py
```diff
@@ -133,6 +133,14 @@ def send_with_options(self, *, args=None, kwargs=None, delay=None, **options):
         Returns:
           Message: The enqueued message.
         """
+        for name in ["on_failure", "on_success"]:
```
Bogdanp: It might make sense to only run this block if the `Callbacks` middleware is set on `self.broker`.

Issue #29, opened 2018-01-09
Task composition (ie: Celery Canvas)


aequitas:
What is your viewpoint and this projects direction in regards to task composition?

I currently use Celery because it allows me to work with tasks on a high level (chain, group, etc). I know Dramatiq supports callbacks but for as far I can see they are rather limited. 

Currently with Dramatiq I cannot trivially:

- Chain more than 2 Actors.
- Pass arguments to a chained Actor during composition (unless I pass them through the first task).
- Composite a more complex task/group and pass it to some other part of my application to have it 'send' (like Celery signatures).

I understand this project might be in it's infancy and you might not want to introduce more complexity then needed, but I'm curious about your roadmap for this project and the use cases.

pirate:
Chaining actors would require adding in a whole event loop system that keeps a queue of tasks that are blocked on a preceding task.

While I see the usefulness, personally I think a scheduler event loop is too complex and high-level to warrant being included in dramatiq.  One of the best things about dramatiq is that it's simple and bare-bones, and if you want to build a more complex system on top of it, there's nothing stopping you.  The requirements for chaining tasks will be so different for different users of this library, I don't see how it can be trivially implemented in a way that suits everyone.  There are so many edge cases around failing tasks, task order dependency, distributing it across multiple workers, what to do when dramatiq goes down halfway through a chained task.

Of course it's up to @Bogdanp, if he thinks there's a simple way to do this safely, I'm all ears.

Bogdanp:
@aequitas can you give a few examples of your use cases? As concrete as possible would be helpful. IME, celery's canvas features end up being a maintenance burden long term and they're generally not very stable (depending on the backend -- I've talked about this elsewhere, but my experience with them has been pretty poor, esp. wrt chord performance and dropped chords).

I agree with @pirate that something like this wouldn't belong in core and can definitely be implemented as an external package.

aequitas:
Thanks for the quick response, just to emphasise I agree with keeping the complexity low. Since you do not want to end up with a second Celery, that'd be pointless. The benefits that Celery brings also has its downsides so it's all about being able to choosing the right tool for the right job.

I think I saw someone on HN mention something like 'Django is to Flask, like Celery is to Dramatiq'. Which is exactly the feeling I had when discovering this project. In that you focus on doing one thing very well and by that produce a clean and lean solution.

[...]

Issue #30, opened 2018-01-13
Task states


mjdubell:
When running multiple tasks you often want to know about the tasks state or show some kind of progress to an end user. I believe it would be beneficial for Dramatiq to include such functionality, just like Celery does, which you can read more about [here](http://docs.celeryproject.org/en/latest/userguide/tasks.html#task-states).

If Dramatiq includes some kind of chain or group functionality, it would be great to get the current state of the chain/group.

ericgj:
Check out current discussion going on now in #29

ericgj:
Bogdan may have plans to integrate this into results storage, but in the meantime you can roll your own using the Callbacks middleware plus a similar middleware that fires on `before_process_message`. 

As for pipeline/groups,  dramatiq doesn't attach any extra state to these - right now you have to attach some kind of pipeline/group id to each message.

What I am doing is prepending a message to the front of my pipelines with the entire rest of the pipeline as the payload (args). This message when executed inserts records for each of the messages in the pipeline with a NOT STARTED status. Then as the messages run they update this status, save results, errors, etc. (You could do it insert-only instead.)  So far it is working well. If you are interested I can share my code in a gist.

Bogdanp:
I'm going to close this as "won't do" since this is easy to implement on top of the existing primitives, like Eric mentioned.

albasili:
@ericgj I would be interested in having a look at your code if that is not much of a trouble.

ericgj:
Well, my approach is a bit "1.0" -- it works for me at the moment but I'm not sure I'd recommend it wider :)  But it may give you some ideas for improvement.

You can see all the code [here][code].  Here is the [results middleware][results_middleware] I use to store results, and customized [pipelines middleware][pipelines_middleware]. Here is how I [construct pipelines][construct] to work with these middlewares, and here is the [message broker setup][broker] showing how success and failure callbacks are specified. Then [here is the actual storage implementation][storage], using SQLite at the moment.

Just to note, you may find a way of building on the results backend primitives built-in to Dramatiq, e.g. through `Message#get_result`. This would no doubt be less workaround-ish. I couldn't quite wrap my mind around how to do it with pipelines and needing to report on individual task status within a pipeline.

[code]: https://bitbucket.org/ericgj/ert-cdp-task-queue/src/
[results_middleware]: https://bitbucket.org/ericgj/ert-cdp-task-queue/src/master/backend/dramatiq_extra/results/middleware.py
[...]

PR #31, opened 2018-01-13
feature: add support for groups and pipelines

Bogdanp:
This is a WIP based on the discussion in #29.


dramatiq/middleware/pipelines.py
```diff
@@ -0,0 +1,37 @@
...
+        # from broker -> pipelines -> messages -> broker.
+        from ..message import Message
+
+        if exception is not None:
+            return
+
+        message_data = message.options.get("pipe_target")
+        if message_data is not None:
+            next_message = Message(**message_data)
+            if not message.options.get("pipe_ignore", False):
```
ericgj: Thanks for this option. In my situation with pipelines constructed by less technical users, and more or less stand-alone, "fire-the-missiles" tasks, chaining results is mostly not needed.

aequitas:
I like the feature to specify different options per message in the pipeline. I think the 'message_with_options' was added in a later iteration (or I just missed it when reading the PR earlier)? 
We have a use case where the same tasks needs to be routed to different workers (due to availability of IPv6 on that worker or not), this feature allows for a nice solution for this.

docs/source/cookbook.rst
```diff
@@ -44,6 +44,103 @@ time an actor fails, even if the message is going to be retried.
...
+
+  pipe.get_result(block=True, timeout=5_000)
+
+To get the intermediate results of each step in the pipeline, you can
+call |pipeline_get_results|::
+
+  for res in pipe.get_results(block=True):
+    ...
+
+
```
bendemaree: WDYT about specifically covering exception handling? I tried it out with a pipeline and found it entirely intuitive (`raise` in a pipeline step and subsequent steps do not execute) but perhaps it's worth mentioning?

Bogdanp: I think that's something we can cover in the future, if there's a need. Result storage only supports results, not exceptions right now so we'd have to expand that scope, too. Might not even be necessary with #30.

PR #61, opened 2018-04-10
Change dependency handling

### LINKED ISSUE #60: dramatiq dependencies
I noticed that by default, dramatiq installs the dependencies for rabbitmq and the watch feature. I think this is due to the following lines:

https://github.com/Bogdanp/dramatiq/blob/f6cea0d3c22a873e949fcc6f9e45b9b626841552/setup.py#L44-L47

Executing `pip install dramatiq[redis]` will still install in the rabbitmq and watch dependencies.

rpkilby:
Should fix #60

- Makes prometheus an extra dependency
- Don't always install rabbitmq+watch dependencies

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -15,3 +15,4 @@ of those changes to CLEARTYPE SRL.
 | [@chen2aaron](https://github.com/chen2aaron) | xixijun |
 | [@aequitas](https://github.com/aequitas) | Johan Bloemberg |
 | [@najamansari](https://github.com/najamansari) | Najam Ahmed Ansari |
+| [@rpkilby](https://github.com/rpkilby) | Ryan P Kilby |
--- a/setup.py
+++ b/setup.py
@@ -41,10 +41,6 @@ for extra in extras:
     extra_dependencies[extra] = list(parse_dependencies(filename))
 
 extra_dependencies["all"] = list(set(sum(extra_dependencies.values(), [])))
-extra_dependencies[""] = list(set(
-    extra_dependencies["rabbitmq"] +
-    extra_dependencies["watch"]
-))
 
 setup(
     name="dramatiq",


docs/source/installation.rst
```diff
@@ -29,14 +29,15 @@ Extra Requirements
...
-``watch``      Installs the required dependencies for the ``--watch`` flag.
-=============  =======================================================================================
+==============  =======================================================================================
+Name            Description
+==============  =======================================================================================
+``memcached``   Installs the required dependencies for the Memcached rate limiter backend.
+``rabbitmq``    Installs the required dependencies for using Dramatiq with RabbitMQ.
+``redis``       Installs the required dependencies for using Dramatiq with Redis.
+``watch``       Installs the required dependencies for the ``--watch`` flag.
+``prometheus``  Installs the required dependencies for the Prometheus statistics middleware.
```
Bogdanp: Prometheus can't be an optional dependency right now because it's a default middleware

Bogdanp:
My intention with the default extras was for `pip install dramatiq` to install RMQ and watch deps. I didn't realize it had the side-effect it had. I wonder if there's a way of doing that w/o the side-effects though?

docs/source/installation.rst
rpkilby: Should prometheus be made non-default, or should this change be reverted? As a user, the former would be preferable since we don't use prometheus.

rpkilby: Although I understand if you'd rather make that change later.

Bogdanp:
That's unfortunate. I just checked the docs briefly and it looks like they always recommend installing w/ explicit extras so this looks good to me! I'll merge it tonight along w/ the other PR. I expect I'll do a 1.1 release sometime next week.

Bogdanp:
Thanks again! I merged your changes with rebase here:

https://github.com/Bogdanp/dramatiq/commit/8a2295cc50fd661e9cd99342be67161ea7d4fdd2
https://github.com/Bogdanp/dramatiq/commit/35d86f7ec543c2255f74f0ef0e35dbc7b33742fe

PR #66, opened 2018-04-11
Signal processing improvement

2miksyn:
According `python` docs `Python signal handlers are always executed in the main Python thread`, but sometimes this is not true. This problem is too old, and on some OSs it periodically take place.
FreeBSD 10.3 amd64 send signal for last python thread, and it is not processed. FreeBSD 11.0 is not affected. Problem occurs on python 3.6 and 3.8.
Blocking signal processing in thread resolves problem independently of OS.

### DIFF
--- a/dramatiq/__main__.py
+++ b/dramatiq/__main__.py
@@ -332,6 +332,8 @@ def main():  # noqa
     def watch_logs(worker_pipes):
         nonlocal running
 
+        signal.pthread_sigmask(signal.SIG_BLOCK, {signal.SIGINT, signal.SIGTERM, signal.SIGHUP})
+
         log_file = args.log_file or sys.stderr
         selector = selectors.DefaultSelector()
         for pipe in [parent_read_pipe] + worker_pipes:


Bogdanp:
So this will block the log watcher thread from receiving signals, correct? What about the filesystem changes watcher then? I would assume that that needs a similar block set put in place.

Bogdanp:
Would a better approach be to block all signals before spawning the child threads and then unblock then in the main thread before setting up the signal handlers?

2miksyn:
Yes, this will block signal processing in log watcher thread. In this case signals always will be processed in main thread. Currently `watch` options not works. I did not study this problem. It may be problem of `watchdog` module. In this case this problem be hard - `watchdog` module has not been updated for 3 years.

2miksyn:
Your approach will work too, but, think, this is not good practice - code will become harder. Think, pass signals at start time is not big problem, I did not fell it in my practice.

2miksyn:
Your approach will work too, but, think, this is not good practice - code will become harder. And I did not see any advantages - signals will be passed at start time too.

Bogdanp:
The problem is the main thread can spawn many threads and I don't want to litter each thread with signal blocking code. I believe the blocking-then-unblocking approach is actually better from a code quality perspective. 

Please try this branch out and let me know if it works for you: https://github.com/Bogdanp/dramatiq/tree/signal-blocking

PR #69, opened 2018-04-15
Recompute keys in atomic incr with sliding window

gdvalle:
Change the `WindowRateLimiter` to pass a callable to `incr_and_sum` so
that during atomic increments the time windows can be recomputed in the
event of a lock.

This breaks the API because we change `incr_and_sum`'s `keys` arg to
a callable. We keep backwards compatibility by checking if `keys` is
callable and calling it on each use.

### DIFF
--- a/dramatiq/rate_limits/backends/memcached.py
+++ b/dramatiq/rate_limits/backends/memcached.py
@@ -100,7 +100,9 @@ class MemcachedBackend(RateLimiterBackend):
                 if value > maximum:
                     return False
 
-                mapping = client.get_multi(keys)
+                # TODO: Drop non-callable keys in Dramatiq v2.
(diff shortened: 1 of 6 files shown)


dramatiq/rate_limits/backends/memcached.py
```diff
@@ -100,6 +100,8 @@ def incr_and_sum(self, key, keys, amount, maximum, ttl):
                 if value > maximum:
                     return False
 
+                # TODO: Drop non-callable keys in Dramatiq v2.
+                keys = keys() if callable(keys) else keys
```
Bogdanp: This should be renamed to `key_list` or something since it's in a loop to avoid shadowing the `keys` callable.

Issue #37, opened 2018-02-19
Cancel long running tasks


larsrinn:
Hi there,
is it possible, to revoke enqueued and cancel long running tasks? For example based on the message id used for starting the task? If not, are there any plans in introducing this in a future version?
Cheers
Lars

Bogdanp:
Long-running tasks can be given time limits. For example:

```python
@dramatiq.actor(time_limit=60_000)
def some_actor():
  pass
```

`some_actor` would be killed after 60 seconds (with some exceptions, eg. if it's blocking on a system call).

Apart from that, there is no way to remotely tell a worker to stop executing a task and it's unlikely that I'll ever introduce something like that. The upcoming dashboard feature will let you mark tasks for cancellation, but those cancellations will only take effect if they are registered before the task runs, so it's not quite the same as what I think you're asking for.

larsrinn:
Yes, I saw the possibility to set a time limit.

So your suggested feature to mark tasks to be canceled is comparable to what celery calls revoke?

I just looked at the implementation of the TimeLimit middleware. Without digging too deep, I guess canceling the task if they are already started, should be possible in a similar manner. Why do you think an introduction will be unlikely? Don't you want to bloat dramatiq or is it a question of available time and other priorities?

Bogdanp:
> Don't you want to bloat dramatiq or is it a question of available time and other priorities?

A little bit of both! It's certainly possible to do, but I would like to avoid doing it until more people request such a feature. Additionally, while stopping the tasks is relatively easy, what's harder is implementing a way to talk to the workers directly that works w/ both RabbitMQ and Redis.

For now, if you absolutely need this, you could define a middleware similar to this:

(Redis)
```python
REMOTE_TOPIC = "dramatiq-remote"

class Controller:
  def __init__(self, client):
    self.client = client
  
  def cancel(self, message_id):
    self.client.publish(REMOTE_TOPIC, json.dumps({"command": "cancel", "message_id": message_id}))

class Subscriber(Thread):
  def __init__(self, pubsub, threads_by_message):
    super().__init__(daemon=True)
    self.pubsub = pubsub
    self.pubsub.subscribe(REMOTE_TOPIC)
    self.threads_by_message = thread_by_message

  def run(self):
    for message in self.pubsub.listen():
      data = json.loads(message["data"])
      if data["command"] == "cancel":
        self._perform_cancel(data["message_id"])

  def _perform_cancel(self, message_id):
    try:
[...]

PR #71, opened 2018-04-20
Worker shutdown notification middleware

rpkilby:
Adds a middleware that notifies actors when their worker process has been signaled for termination. Like the `TimeLimit` (and proposed task cancellation) middleware, an exception is raised in the actor so that the code can properly clean up after itself. Actors opt into the behavior by setting the `notify_shutdown` option to True. If the option is set, the `Shutdown` exception is raised in the actor whenever its worker process is sent a sigterm. If the option is not set, then the exception is not raised.

This PR consists of two chunks:

1) Factor out the exception raising behavior of `TimeLimit` into a new `ThreadExceptingMiddleware` 
    - (might also be useful for the proposed `RemoteControlMiddleware` from #37).
2) Add the `ShutdownNotifications` middleware, which inherits this new middleware base class. 

At this point the PR is mostly ready, but I wanted to make sure this was worth pursing before cleaning it up.

**Possible todos:**
- Review naming. eg, is `ThreadExceptingMiddleware` ideal?
- more tests
- add docs
- add example


Bogdanp:
Took a quick peek at this and it looks great! I'll do a more in-depth review tonight.

dramatiq/middleware/shutdown.py
```diff
@@ -0,0 +1,69 @@
...
+        actor = broker.get_actor(message.actor_name)
+
+        if message.options.get("notify_shutdown") or actor.options.get("notify_shutdown"):
+            self.notifications.add(threading.get_ident())
+
+    def after_process_message(self, broker, message, *, result=None, exception=None):
+        thread_id = threading.get_ident()
+
+        if thread_id in self.notifications:
+            self.notifications.remove(thread_id)
```
Bogdanp: You should add `after_skip_message = after_process_message` underneath this.

dramatiq/middleware/threading.py
```diff
@@ -0,0 +1,67 @@
...
+
+import ctypes
+import platform
+import warnings
+
+from ..logging import get_logger
+from .middleware import Middleware
+
+
+class ThreadExceptingMiddleware(Middleware):
```
Bogdanp: I think it might be better not to rely on inheritance here. Just extract `raise_thread_exception` as a normal function in this module that the middleware can use. We can leave out the platform warning or we can have each of the middleware implement it individually, I don't see much value in DRYing that up, especially if it means having to follow the inheritance tree up in order to figure out where it's firing from.

dramatiq/middleware/shutdown.py
rpkilby: Should this also apply to the `TimeLimit` middleware?

Bogdanp: Yes, it should. Looks like I missed it when I added `SkipMessage`.

rpkilby:
Hi @Bogdanp. I've updated the code and rebased/reworked the commit history. The code changes should be ready for review, although I'm still working on updating the docs and adding an example.

- Thread raising reworked from base class to function
- Added base `Interrupt` exception for `TimeLimitExceeded` and `Shutdown`. In general, this seems like a good term to describe the behavior. These middleware are asynchronously raising an exception on worker threads in order to interrupt the actor's execution. 
- Added `tests/middleware` to house the threading and shutdown tests. I've moved the prometheus tests into there as well. Does that work, or do you prefer having them all in a flat tests directory?
- Added `ShutdownNotifications to the `default_middleware`, and moved `default_middleware` into the `middleware` package.

dramatiq/middleware/shutdown.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def after_process_boot(self, broker):
+        if current_platform not in supported_platforms:  # pragma: no cover
+            msg = "ShutdownNotifications cannot kill threads on your current platform (%r)."
+            warnings.warn(msg % current_platform, category=RuntimeWarning, stacklevel=2)
+
+    def before_worker_shutdown(self, broker, worker):
+        self.logger.debug("Sending shutdown notification to worker threads...")
+        for thread_id in self.notifications:
+            self.logger.info("Worker shutdown notification. Raising exception in worker thread %r.", thread_id)
```
rpkilby: Is it relevant to include the `thread_id` here, since the worker's log format includes the thread name?

rpkilby:
Alright, added docs and an example. Everything should be ready for review now.

dramatiq/middleware/shutdown.py
Bogdanp: I believe so, because the thread logging the message may be different from the thread executing the task.

Bogdanp:
Thanks again! I merged this with rebase here: https://github.com/Bogdanp/dramatiq/compare/d323bb027be6187dafea793e3010a40a0fe4eebf...5c21bb7

PR #72, opened 2018-04-22
Make test suite more friendly to new contributors

rpkilby:
Hi @Bogdanp - I ran into a few issues when I started writing tests. Namely, running the test suite requires having all of the services configured correctly, which may not be the case. As a result, the test suite fails outright. If you delete the benchmarks (which cause the outright failure), the tests still result in ~80 errors due to connection failures. I've made the following changes:

- In dev, skip tests when the required rabbitmq/redis/memcached service is not available.
- Don't run the benchmarks as part of the default test suite. They're take a fair amount of time, and most users will not want to wait for them to complete. 
- Deleted a test fixture for the old url-based RabbitMQ broker

I've also made the following changes to the travis config:
- Run flake8 in its own CI job.
- Run the benchmarks on one py36 job, instead of on all of version. This can be reverted - it wasn't entirely clear to me if it was important to run the benchmarks on all versions of python.

Thoughts?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,10 @@
 import dramatiq
 import logging
+import os
+import pylibmc
 import pytest
 import random
+import redis
 import subprocess
 import sys
 
@@ -17,6 +20,29 @@ logging.basicConfig(level=logging.INFO, format=logfmt)
 logging.getLogger("pika").setLevel(logging.WARN)
 
 random.seed(1337)
+
+CI = os.environ.get("TRAVIS") == "true"
+
+
+def check_rabbitmq(broker):
+    try:
+        broker.connection
+    except Exception as e:
+        raise e if CI else pytest.skip("No connection to RabbmitMQ server.")
+
+
+def check_redis(client):
+    try:
+        client.ping()
+    except redis.ConnectionError as e:
+        raise e if CI else pytest.skip("No connection to Redis server.")
+
+
+def check_memcached(client):
+    try:
+        client.get_stats()
+    except pylibmc.SomeErrors as e:
+        raise e if CI else pytest.skip("No connection to memcached server.")
 
 
 @pytest.mark.hookwrapper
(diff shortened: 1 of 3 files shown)


rpkilby:
I just noticed that the benchmarks aren't actually ran on travis. Would it make sense to start doing this, given the separate build?

tox.ini
```diff
@@ -10,20 +11,22 @@ deps=
...
 passenv=
   TRAVIS
 
+[testenv:benchmarks]
+deps =
+  -r{toxinidir}/requirements/dev.txt
+install_command=
+  pip install -U -c {toxinidir}/requirements/constraints.txt {opts} {packages}
+commands=
+  py.test --ignore=tests tests/benchmarks {posargs}
```
Bogdanp: You can use `py.test --benchmark-only` here.

tox.ini
```diff
@@ -10,20 +11,22 @@ deps=
   pypy: -r{toxinidir}/requirements/dev-pypy.txt
   flake8: flake8
 commands=
-  py.test {posargs}
+  py.test --ignore=tests/benchmarks {posargs}
```
Bogdanp: You can use `py.test --benchmark-skip` here.

tests/conftest.py
```diff
@@ -18,6 +21,29 @@
 
 random.seed(1337)
 
+CI = os.environ.get('TRAVIS') == 'true'
```
Bogdanp: Nitpick: use double quoted strings for consistency.

Bogdanp:
> I just noticed that the benchmarks aren't actually ran on travis. Would it make sense to start doing this, given the separate build?

I think it would be worth running them on travis is we could somehow preserve the results between builds so that we could fail builds in case of a regression, but I don't know if that's possible using Travis.

Bogdanp:
Thanks! Merged with rebase here: https://github.com/Bogdanp/dramatiq/compare/5c21bb74175c0d1c73d07046509d5f35e1c4d3da...e79ce8b

PR #73, opened 2018-04-22
reformat all code using isort and flake8-quotes

Bogdanp:


dramatiq/__init__.py
```diff
@@ -33,33 +38,40 @@
 
 __all__ = [
     # Actors
-    "Actor", "GenericActor", "actor",
-
+    "Actor",
```
rpkilby: You may want to consider one of the [multi line output modes](https://github.com/timothycrosley/isort#multi-line-output-modes), such as `multi_line_outuput=5`

Bogdanp: I was just looking at that but I don't think it applies to `__all__`. Also, I think the only mode that's compatible between black and isort is `3`.

rpkilby: Just to further throw in my two cents - I have mixed feelings about black's output. There are several style changes that seem like an improvement, but more that seem like they hurt readability. eg, changes to the argument parser, several method and test signatures, etc... idk if this is a result of black trying to format on a shorter line length.

¯\\\_(ツ)_/¯

Bogdanp: Yeah, I think I might drop black for now since it does hurt readability and it doesn't support Python 3.5.

rpkilby: In the mean time, flake8 has a number of extensions. eg, double quote enforcement could be handled with https://github.com/zheller/flake8-quotes

Bogdanp: Nice! I wasn't aware of that.

PR #76, opened 2018-04-27
Add Windows CI support (POC)

2miksyn:
It is only POC. Has problem with installing memcached and pylibmc. Thinks, it will help to develop and test on Windows in future.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,74 @@
+environment:
+  matrix:
+  #  - PYTHON: "C:\\Python35"
+  #    PYTHON_VERSION: "3.5.x"
+  #    PYTHON_ARCH: "32"
+
+    - PYTHON: "C:\\Python35-x64"
+      PYTHON_VERSION: "3.5.x"
+      PYTHON_ARCH: "64"
+
+  #  - PYTHON: "C:\\Python36"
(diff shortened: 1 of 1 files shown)


Bogdanp:
Thanks! Any indication why memcached and pylibmc don't work?

2miksyn:
Currently you can see problem with installing pylibmc in your appveyor account. But is not all. If you uncomment strings in appveyor.yml connected with memcached (server), you can see, what choco can not install it. And install logs not details problem properly. I found some recipe: fetch zip with Windows memcached build from some site and then manually unpack and start it (this recipe use PHP developers). Thinks, what it is choco bug and better way - resolve it and install with choco.
If comment pylibmc in requirements/dev.txt, then build starts and you can see fnctl problem in tests.
Problem with installing pylibmc connected with dependence of libmemcached (or, as I think with libmemcached-dev). Thinks, decision is manually fetch it from any site and unpack.
Maybe simple way - use another python module (not pylibmc) on Windows. But, thinks, install libmemcached is not really problem.
I can not time for resolving problems, but may be I will return to them.

Bogdanp:
Thanks again! Merged here: https://github.com/Bogdanp/dramatiq/commit/6667a686080156513003b480e368958bdce19f2a I'll try to iterate over it until I get it working.

Issue #78, opened 2018-04-28
Get AppVeyor build working


Bogdanp:
There's a partially-working `appveyor.yml` in the repo, but someone that has access to a Windows machine (and a whole heap of patience) needs to get it across the finish line. Ideally, someone would get it up the point where the tests run (and fail due to https://github.com/Bogdanp/dramatiq/issues/75). 

Current state:

* pylibmc fails to build against libmemcached due to missing dependencies and, seemingly, lack of c99  support

Bogdanp:
Thanks to @ryansm1 for getting this done!

PR #112, opened 2018-09-26
Appveyor config

ryansm1:
This is to address issue #78. As expected, the build will fail at the onset of testing because fcntl is missing.

Notes:
- libmemcached is compiled without SASL support
- pylibmc is compiled without zlib support


Bogdanp:
Thanks, @ryansm1! Did you intend to close this?

ryansm1:
@Bogdanp yes, sorry. I realized that the appveyor config had issues even though it failed as expected here—this one wasn't starting RabbitMQ/Redis and so the tests would still fail despite fcntl/fork being unavailable. I'll send another PR once I have it fixed.

PR #113, opened 2018-09-26
Working appveyor configuration

ryansm1:
This is to address issue #78. As expected, the build will fail at the onset of testing because fcntl is missing. This is a replacement to PR#112 and ensures that Redis is started after installation.

Notes:
- libmemcached is compiled without SASL support
- pylibmc is compiled without zlib support


Bogdanp:
Huge thank you for tackling this! I've merged your changes (and added more on top) in https://github.com/Bogdanp/dramatiq/commit/3f983202bf2e90e156c82dcc95a155353eb63f02 and https://github.com/Bogdanp/dramatiq/commit/d470d4971574fa73231f7175544cfc25756fcfcb .

PR #115, opened 2018-10-06
Use Windows-compatible libmemcached source

ryansm1:
This uses a modified libmemcached from @yshurik to build a working pylibmc. I made changes to the VS project file by hand to allow it to compile under x86 or x64 architectures.

The previous configuration successfully compiled libmemcached but pylibmc could not connect to a memcached server. This build works on Windows using the MSVC compiler.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-image: Visual Studio 2015
+image: Visual Studio 2017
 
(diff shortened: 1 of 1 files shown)


Bogdanp:
Awesome! Thanks! :D

Looks like we're down to 8 failing tests now which is fantastic.

PR #118, opened 2018-10-06
Fixing docs build

aericson:
`sphinxcontrib-versioning` does not work with sphinx 1.8.x.

Pinning version to be able to build the docs.

You can see the build error here: https://travis-ci.org/Bogdanp/dramatiq/jobs/437987961


Bogdanp:
Thanks! I merged your changes with rebase.

PR #119, opened 2018-10-09
Windows compatibility

ryansm1:
Summary of changes:
- Move code from ``__main__.py`` into ``main.py`` to prevent picklability errors on Windows when spawning new processes (diff ``main.py`` with ``__main__.py`` to see line-by-line edits)
- Use multiprocessing's Process to spawn worker processes instead of os.fork()
- Remove TimeLimit middleware on Windows platforms
- Use logging's QueueHandler instead of pipes for interprocess logging
- Setup SIGHUP (POSIX) and SIGBREAK (Windows) signal handlers only if they are available
- Skip tests that cannot pass on Windows (e.g. tests requiring signals)

Known issues:
- ~~Gevent script is broken on Windows~~


ryansm1:
Hopefully the additional commits don't interrupt your review; I wanted to clean up a couple of things that I missed after my last round of changes. It looks like some of the RabbitMQ retry-after-delay tests fail intermittently.

These are pretty substantial modifications to the main script so I expect them to need some additional work; let me know how I can help!

Bogdanp:
Thanks again for doing this! I've reviewed your changes and made some of my own on top in https://github.com/Bogdanp/dramatiq/pull/126 . Please take a look and let me know what you think. 

I'm going to sit on that PR for a bit because it breaks gevent support (as mentioned in one of your commits) and disabling thread patching is not acceptable since the reason you're likely to run gevent with dramatiq is because you want to run many worker threads concurrently. [This PR](https://github.com/gevent/gevent/pull/1158) seems like it may fix the issue but there's no real indication of when it's going to be ready so we might have to find a different solution in the mean time (maybe `multiprocessing.Pipes`?).

PR #120, opened 2018-10-12
Declare queue lazily

etiennecrb:
Currently, a queue is declared when an actor is. This means a connection
to broker must be established when actors are imported. It leads to some
difficulties when you need to change some configuration parameters after
actors have been imported.

This change declare queue lazily, meaning on first message enqueuing.


etiennecrb:
Oops! This PR was opened mistakenly :wink:

PR #126, opened 2018-10-21
Enhancements+windows compat

Bogdanp:
Todos:

* [ ] Add more info re. Windows support to docs
* [ ] Appveyor build is still a little flaky 
* [ ] Figure out a workaround for gevent deadlock with `multiprocessing.Queue` (disabling thread patching is unacceptable) or wait for https://github.com/gevent/gevent/pull/1158


Bogdanp:
@ryanhiebert this is mostly ready to go. It just needs a little more testing so you can help with that if you like. I'd make sure all the examples still run as expected w/ every broker combo and under gevent. I plan on doing that this weekend or the next (depending on how much time I have).

Another thing that would be nice to get done before this is merged, but I don't think should block it is the AppVeyor build flakiness. If you've got a Windows machine and would like to take a stab at that, I'd def appreciate it!

PR #130, opened 2018-10-28
Pass arguments explicitly so we can override it

maerteijn:
With this pull request it becomes possible to write a custom entrypoint with a different argument parser or another source of arguments (for example a config file). Now this is almost impossible without monkey patching or duplicating a lot of code.

A custom entrypoint then looks like this:

```python
import sys

from dramatiq.__main__ import argument_parser, main


def entrypoint():
    args = ['my.module:broker', 'my.module.actors', '-Q', 'myqueue']
    parsed_args = argument_parser().parse_args(args)
    return main(parsed_args)

if __name__ == "__main__":
    sys.exit(entrypoint())
```

(ps, and thanks for this awesome piece of software)


maerteijn:
(ps, the travis / appveyor builds fail, but it looks like this pull request has nothing to do with it right?)

maerteijn:
Hi @Bogdanp Any chance you have a look at this? This would be really helpful for a couple of usecases in my projects. Thanks.

Bogdanp:
Hi @maerteijn. I will be merging this after #126, but that may take another week or two as I need to find time to properly test it.

maerteijn:
I see that you changed a lot of the code structure in #126 Shall I update my pull request to be rebased on that branch so it can be merged?

maerteijn:
I adjusted the pull request so it can be merged into your `enhancements+windows-compat` branch. 

I also pinned the `redis` package to `<3.0.0` as there are some backwards incompatible changes (https://github.com/andymccurdy/redis-py/blob/master/CHANGES#L27 for example).
Compatibility with  the `redis>=3.0.0` should be something to fix in another pull request I think.

The same goes for `pylibmc`, I pinned it on `<1.6` as the `test_concurrent_rate_limiter_releases_the_lock_after_each_call[memcached]` test fails.

Bogdanp:
Whoops, looks like I accidentally closed this by removing the `enhancements+windows-compat` branch. Would you mind rebasing on top of master and re-opening the PR?

maerteijn:
@Bogdanp I created a new pull request which is rebased on master, see #140

PR #140, opened 2018-11-24
Parse arguments outside of main so we can have a custom entrypoint

maerteijn:
With this pull request it becomes possible to write a custom entrypoint with a different argument parser or another source of arguments (for example a config file). Now this is almost impossible without monkey patching or duplicating a lot of code.

A custom entrypoint could looks like this:

```python
import sys

from dramatiq.cli import argument_parser, main


def my_entrypoint():
    args = ['my.module:broker', 'my.module.actors', '-Q', 'myqueue']
    parsed_args = argument_parser().parse_args(args)
    return main(parsed_args)

if __name__ == "__main__":
    sys.exit(my_entrypoint())
```

Changing the existing arguments or adding extra ones will be much simpler then as well. Could also simplify the `dramatiq-django` integration by calling the python module directly instead of `os.execvp`.

### DIFF
--- a/dramatiq/cli.py
+++ b/dramatiq/cli.py
@@ -118,7 +118,7 @@ def folder_path(value):
     return os.path.abspath(value)
 
 
-def parse_arguments():
+def argument_parser():
     parser = argparse.ArgumentParser(
         prog="dramatiq",
         description="Run dramatiq workers.",
@@ -177,7 +177,7 @@ def parse_arguments():
 
     parser.add_argument("--version", action="version", version=__version__)
     parser.add_argument("--verbose", "-v", action="count", default=0, help="turn on verbose log output")
-    return parser.parse_args()
+    return parser
 
 
 def setup_pidfile(filename):
@@ -322,8 +322,7 @@ def worker_process(args, worker_id, logging_pipe):
     logging_pipe.close()
 
 
-def main():  # noqa
-    args = parse_arguments()
+def main(args):  # noqa
     for path in args.path:
         sys.path.insert(0, path)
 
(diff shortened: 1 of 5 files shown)


Bogdanp:
Thanks! Please make a commit adding yourself to the contributors file.

dramatiq/cli.py
```diff
@@ -118,7 +118,7 @@ def folder_path(value):
     return os.path.abspath(value)
 
 
-def parse_arguments():
+def argument_parser():
```
Bogdanp: Is there a particular reason to change this function? Do you plan to augment the argument parser?

maerteijn: Wlel, the function first parsed the arguments and returned them. Now it returns the argument parser itself. 

It think it should be clear what the function is doing, that's why I suggest to change the name.

Bogdanp: Right, I wasn't asking about the name specifically. I was asking about the reasoning behind changing it to return the parser. :D

maerteijn: Oh, sorry for the misunderstanding 😃 I have two existing scenarios in our project where we would like to pass the arguments to the main function instead of parsing the arguments inside it.

1) A custom entrypoint (`console_script`) with predefined arguments, see the example in the description of this pull request
2) A custom entrypoint (`console_script`) which does not parse the arguments from the commandline but from a configuration file deployed with our configuration manager

Now I have to mokeypatch the `main` function in `cli.py` to do so because there the parsing happens.

Bogdanp:
Thanks! I made a couple small changes and merged this with rebase. I'll cut a release later today.

Issue #136, opened 2018-11-24
Upgrade to redis-py 3.0


Bogdanp:
The latest version of the redis client made a number of [breaking changes](https://github.com/andymccurdy/redis-py/blob/master/CHANGES) that affect us.  For now, I've pinned it to the 2.x series, but we'll eventually have to upgrade.

Bogdanp:
Done in #142.

PR #142, opened 2018-11-27
WIP: Support redis py 3.0

maerteijn:
As I saw the "Help wanted" label for #136 here my first attempt (WIP) to fix the issues so Dramatiq works with the redis-py 3.0 package. 

For as far I can see from the [breaking changes](https://github.com/andymccurdy/redis-py/blob/master/CHANGES) the following issues have impact on dramatiq:

* The StrictRedis class has been renamed to Redis. StrictRedis will continue to exist as an alias of Redis for the forseeable future.
* The legacy Redis client class has been removed. It caused much confusion to users.
* ZADD now requires all element names/scores be specified in a single dictionary argument named mapping. This was required to allow the NX, XX, CH and INCR options to be specified.

These issues are addressed in this pull request. `disptach.lua` seems to work fine, however I'm not so a LUA programmer so it could be luck that it still works 😄 Will investigate later or maybe you have some hints for me

### DIFF
--- a/dramatiq/brokers/redis.py
+++ b/dramatiq/brokers/redis.py
@@ -60,7 +60,7 @@ class RedisBroker(Broker):
       >>> RedisBroker(url="redis://127.0.0.1:6379/0")
 
     See also:
-      StrictRedis_ for a list of all the available connection parameters.
+      Redis_ for a list of all the available connection parameters.
 
     Parameters:
       url(str): An optional connection URL.  If both a URL and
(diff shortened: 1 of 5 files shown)


Bogdanp:
Thanks! I'll merge this for `1.5` sometime in the next couple of months.

Bogdanp:
Thanks again! I merged your changes with rebase (and some of my own tweaks on top) and will cut a release later today.

Issue #103, opened 2018-07-22
Feature request: create pipeline of groups


AndreaOrru:
Currently, it's possible to create a group where the children are pipelines:
`children (Iterator[Message|group|pipeline]) – A sequence of messages, groups or pipelines.`

However, it doesn't seem to be possible to create a pipeline with group as children:
`children (Iterator[Message|pipeline]) – A sequence of messages or pipelines. Child pipelines are flattened into the resulting pipeline.`

Any plan to add this feature? Or any suggestions on how to implement it?

Bogdanp:
This isn't currently possible because groups don't have any notion of a completion callback. i.e. to be able to use a group inside a pipeline you'd have to be able to send a message to the next actor in line a group completes. It may be possible to extend groups to support something like this, but I'll have to think about it more.

Can you tell me about your use case in the mean time?

ryanhiebert:
Here's my use-case for this feature. We have a group where the functions gather some basic data about a bunch of courses at a school's online system. Once it's done, we figure out what users are in those courses and get that data in a callback function. We need to ensure that this happens after that group is done, or else we won't have the full list of users. Our use-case doesn't require the results to be passed along, we're just using it for timing, though I'm sure others' use-cases might want to actually pass the results along.

ryanhiebert:
I've been reading the code for the past few hours to see if I can implement this feature. I've also been reading Celery's code to figure out how they are doing this marshaling with Redis. I don't have a solid plan yet for implementation, but I'm working on it.

I've got to say that this project's code is an absolute pleasure to read. The difference between this code and Celery's is night and day, and it's not just because of the additional functionality that Celery tries to offer. The care that you've taken on this code shows, and it's going to make the job of implementing this feature very targeted and light. This is excellent work. Thank you, @Bogdanp.

ryanhiebert:
As I've read the implementation, it's become clear that, strictly speaking, the result middleware isn't required in order to make the task pipeline work. Celery absolutely requires the result storage in order to make chords (groups with a callback) work. I have an idea to continue that tradition, but I think there is reasonable concern for it to be a bad idea, so I wanted to bring it up, and see if I can get anyone else's thoughts on the approach.

----

First off, I think it's important to have a vision of what Celery does, or at least how I'd envision to implement a similar approach to Celery's in our code. This description is both imprecise and wrong, but close enough to help understand how it works.

[...]

PR #144, opened 2018-12-06
[WIP] Complex Pipelines

ryanhiebert:
References #103

I haven't done any end-to-end testing of this just yet, I've only done some manual spot checking of generated messages to see if they look like what I intend them to look like. I'm not totally sure how this work can or should be integrated into the project, and most of the names I don't really care about, so I'm flexible to rename entities.

This nominally supports arbitrary nesting of three types: Single, Chain, and Group. Single is a wrapper for messages. Chain and Group are for sequential and parallel sets of tasks, respectively. The tasks in a chain or group can be other chains or groups, and conceptually all mixes of them are supported.

However, I haven't yet handled the case of empty groups or empty chains. I expect that empty groups will be something worth having good behavior for, and not cause errors, while empty chains are, IMO, a useless idea and I don't see any reason to support them.

I've created several abstract base classes to help organize my thoughts as I was designing these interfaces. I'm not tied to keeping all those ABCs separate, necessarily, but they were helpful for communicating which functionality is for what purpose.

I'm not tied to the names of all these classes or method names, but I've worked hard on the general design, and I'm pleased with where it's ended up for the most part.

This doesn't currently support passing data to the next task in a chain. My primary objective was to get the flow working, and I believe that it can be addressed after the flow implementation is something that can be reasonable added later. It will involve some extra complexity, because that will also require figuring out how to store and retrieve results for groups. Therefore, I figure it's best to get feedback on the design before I muddy the water trying to bring to many things together at the same time.

One possible improvement, depending on what thoughts we might have for integrating this with the project deeply, would be to allow chains and groups to take messages directly, rather than requiring them to be wrapped in a Single. Leaving the extra wrapping as a requirement was a shortcut to hopefully make this initial review more manageable.

I've written this all with Python 3.6+ in mind, with type annotations. Currently the project support Python 3.5. I'd love to use dataclasses for this work, if we think that could be reasonable, but that would require dropping 3.5 support. What is your plan as far as supporting old versions of Python 3? I don't expect it, but if we wanted to go to Python 3.7-only, I could even make the type signatures a bit neater with a future import. ([PEP 563](https://www.python.org/dev/peps/pep-0563/))

### DIFF
--- a/dramatiq/schemes.py
+++ b/dramatiq/schemes.py
@@ -1,0 +1,261 @@
+from typing import List, Optional
+from abc import ABC, abstractmethod
+from uuid import UUID, uuid4
+import random
+
+from .middleware import Middleware
+from .rate_limits import Barrier
+from .rate_limits.backend import RateLimiterBackend
+from .broker import Broker
+from .message import Message
+
+
+class GroupStart:
+    def __init__(self, *, id: str, count: int):
+        """An object to describe a group to start."""
+        self.id = id
+        self.count = count
+
+    def __repr__(self):
+        return f'GroupStart(id={self.id!r}, count={self.count!r})'
+
+
+class Schemes(Middleware):
+    """Middleware to enable complex task groupings."""
+
+    def __init__(self, backend: RateLimiterBackend, ttl: int):
+        self.backend = RateLimiterBackend
+        self.ttl = ttl
+
+    def barrier(self, group_id):
+        return Barrier(self.backend, group_id, ttl=self.ttl)
+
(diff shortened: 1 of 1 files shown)


Bogdanp:
Thanks for your work on this, but I'm going to close this PR in favor of the approach in #150 because I would like to avoid changing the composition primitives too much.

ryanhiebert:
FTR, my intent with this pull request wasn't to merge as-is, or even close to as-is. It was to provide a starting point for discussion of a design that was capable of all the composition primitive combinations.

PR #150, opened 2018-12-23
feat(composition): add support for group completion callbacks

Bogdanp:
Re. the discussion in #103

### DIFF
--- a/dramatiq/composition.py
+++ b/dramatiq/composition.py
@@ -260,16 +260,32 @@ class group:
                     "group callbacks."
                 )
 
+            # Generate a new completion uuid on every run so that if a
+            # group is re-run, the barriers are all separate.
+            # Re-using a barrier's name is an unsafe operation.
             completion_uuid = str(uuid4())
             completion_barrier = Barrier(rate_limiter_backend, completion_uuid, ttl=GROUP_CALLBACK_BARRIER_TTL)
             completion_barrier.create(len(self.children))
 
             children = []
             for child in self.children:
-                children.append(child.copy(options={
-                    "group_completion_uuid": completion_uuid,
-                    "group_completion_callbacks": self.completion_callbacks,
-                }))
+                if isinstance(child, group):
+                    raise NotImplementedError
+
+                elif isinstance(child, pipeline):
+                    pipeline_children = child.messages[:]
+                    pipeline_children[-1] = pipeline_children[-1].copy(options={
+                        "group_completion_uuid": completion_uuid,
+                        "group_completion_callbacks": self.completion_callbacks,
+                    })
+
+                    children.append(pipeline(pipeline_children, broker=child.broker))
+
+                else:
+                    children.append(child.copy(options={
+                        "group_completion_uuid": completion_uuid,
+                        "group_completion_callbacks": self.completion_callbacks,
+                    }))
         else:
             children = self.children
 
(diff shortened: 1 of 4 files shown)


ryanhiebert:
I am following this PR, I just haven't had time to write up a review yet between holiday happenings. I will get to it as soon as I can.

dramatiq/composition.py
```diff
@@ -224,7 +242,51 @@ def run(self, *, delay=None):
...
+                    "to set it up? It is required if you want to use "
+                    "group callbacks."
+                )
+
+            # Generate a new completion uuid on every run so that if a
+            # group is re-run, the barriers are all separate.
+            # Re-using a barrier's name is an unsafe operation.
+            completion_uuid = str(uuid4())
+            completion_barrier = Barrier(rate_limiter_backend, completion_uuid, ttl=GROUP_CALLBACK_BARRIER_TTL)
+            completion_barrier.create(len(self.children))
```
ryanhiebert: While writing #144, I determined that while the first barrier in a chain of groups should be created immediately, I think that callback groups need to have their barriers created by the middleware. I recognize that your implementation doesn't, yet, support groups as callbacks, but I do think that you'll want to do that.

This is what I created the `groups_starts` option for. And the matching `GroupStart` data class, but that's an implementation detail, it can be whatever form it needs, but it needs to allow for signaling (1) which group to start, and (2) how many tasks are in the group that will start.

In order to support directly nested groups (still raises a `NotImplementedError` in your current implementation, but written in such a way to suggest to me that you may be intending to), you'll need to allow for starting multiple groups at once.

Given the other options you're adding in this PR, I'll suggest that this should be called `group_completion_starts`, and have a shape like this:

```python
message.options['group_completion_starts'] = [
    {'uuid': 'the-first-uuid', 'count': 25},
    {'uuid': 'the-second-uuid', 'count': 50},
]
```

dramatiq/composition.py
```diff
@@ -224,7 +242,51 @@ def run(self, *, delay=None):
...
+                    pipeline_children[-1] = pipeline_children[-1].copy(options={
+                        "group_completion_uuid": completion_uuid,
+                        "group_completion_callbacks": self.completion_callbacks,
+                    })
+
+                    children.append(pipeline(pipeline_children, broker=child.broker))
+
+                else:
+                    children.append(child.copy(options={
+                        "group_completion_uuid": completion_uuid,
```
ryanhiebert: Groups of groups, unlike pipelines of pipelines, cannot be reduced, because there will a functional distinction when results are eventually considered, which I'm sure is an eventual goal of this feature, though I would try to leave it to a later iteration.

To that end, you'll need to make this property take many values, and because of that I'd recommend changing the name to `group_completion_callbacks`, and having it be a list instead of a string. Even if you decide to punt on the group-of-groups feature for now and just leave it `NotImplemented`, I'd encourage changing the name so that we don't need to write compatiblity code that will need to handle multiple names in the future.

ryanhiebert:
I'm continuing work on the design of this, and I'm finding that nested groups are more complicated than the design I espoused above is sufficient to handle. I do think there is an answer, and I'll report back here with it when I've got it figured out, but the eventual, and I think necessary, goal of allowing us to gather results for groups makes this more challenging.

ryanhiebert:
I take that back, I think that is sufficient, as long as zero-child groups are handled carefully, and assuming that the result backend is capable of storing nested arrays in some manner. Sorry for the false alarm.

kurtmckee:
It looks like the branch associated with this PR can be deleted.

PR #152, opened 2018-12-27
fix(actor): optimize the generation of 'actor_name'

wy-z:
Fix the default 'actor_name' to avoid actor name conflict.

Signed-off-by: weiyang <weiyang.ones@gmail.com>


Bogdanp:
Sorry, but this is a breaking change so I can't accept it in 1.x.

wy-z:
It's fine, but the default `actor_name` is easy to cause conflict~
Thanks for your contribution~

Issue #151, opened 2018-12-24
Pipeline continue even if the message is marked as failed


davidt99:
I have a middleware that marks certain messages as failed. The pipeline middleware is only looking whether an exception was raised and continue the chain although I had expected it to stop.
Should be an easy fix, let me know if you want me to open a PR to resolve it (if this is the desired behavior).

PR #153, opened 2018-12-27
fix(pipelines): don't continue when a message marked as failed

davidt99:
fix Bogdanp/dramatiq#151


Bogdanp:
Thanks! I merged this with rebase just now.

PR #155, opened 2019-01-04
WIP: Add arbitrary composition

ryanhiebert:
References #103, #150, #144.

I let myself dwell on the implementation of #144 for a while, and I eventually came up with a simple callable design, implemented here as a singledispatch function, that eliminates the weird cross-referencing of my previous attempt. However, it keeps all of the functionality, and uses essentially the same design for the data to be stored in the message options.

I've rolled all this into a new middleware I called `Composition`, and built the functionality on top of the already existing classes, rather than rolling new ones. (Though they were really helpful in allowing me to understand the basics of the design).

I haven't answered every question about how the current `pipeline` and `group` classes should behave in this new world, but I've answered how the basic mechanics of getting the tasks to run in the right way should work.

For *Right Now*, I've punted on allowing empty groups. I will make that happen, but I wanted to get this into review as soon as I could show anything useful.

I think that the design of the middleware's options is the main thing that I believe is complete for the functionality I'm looking to add in this iteration.

No tests just yet. I'll write those after I get some feedback on the direction that I've got here.

### DIFF
--- a/dramatiq/composition.py
+++ b/dramatiq/composition.py
@@ -16,7 +16,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import time
+import uuid
+from functools import singledispatch
 
+from .middleware.composition import Composition, GroupStart
+from .message import Message
 from .broker import get_broker
 from .results import ResultMissing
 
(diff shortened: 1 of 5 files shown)


ryanhiebert:
@Bogdanp : I'd like to continue putting some time into this, but before I do that I'd like you to take a quick look so you can tell me if I'm way off track here. I've convinced myself that this is a good approach, but I'd like to know if you agree that I'm going in the right direction and that this work is a reasonable basis for starting work and discussion.

Bogdanp:
Ryan, sorry but I plan on finishing this myself in #150. 

I've looked at the code, but unless I can spend a few uninterrupted hours thinking about it, I won't be able to trust that it does everything right and, if I do find enough time to do that, I might as well finish #150 myself.

Issue #91, opened 2018-05-24
Actor's options are ignored


davidt99:
The documentations states that the actor's arbitrary options are passed to the broker and middlewares but that doesn't seem to happen when declaring a queue inside a broker for example.

Bogdanp:
They are only "passed" to those middleware hooks that either take an Actor or a Message. Why do you ask? Is there something specific this is preventing you from doing?

davidt99:
I'm implementing a rabbitmq broker that support rabbitmq's [priority queue](https://www.rabbitmq.com/priority.html). The queue needs `x-max-priority` set when creating the queue. Since the queue name is defined on the actor, I wanted to have this defined on the actor's options as well.

Bogdanp:
It's not ideal (in fact, it's a little ugly), but you can probably achieve this by looking up the actor inside of your `declare_queue` method. I may consider changing the interface for 2.0 to make this sort of thing easier, but I'll leave it as is for now!

PR #157, opened 2019-01-12
RabbitMQ priority

davidt99:
Add support to rabbitmq priority queue.
Examples and documentation are still missing, I'll add them after you'll approve it.
The idea is to send the message with options like this:
```python
some_actot.send_with_options(broker_priority=5)
```

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -63,13 +63,14 @@ class RabbitmqBroker(Broker):
         connection parameters are provided, the URL is used.
       middleware(list[Middleware]): The set of middleware that apply
         to this broker.
+      max_priority(int): Configure the queues with x-max-priority to support priority queue in RabbitMQ itself
       **parameters(dict): The (pika) connection parameters to use to
         determine which Rabbit server to connect to.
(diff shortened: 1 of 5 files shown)


davidt99:
And fixed a bug while writing tests: The worker will start consuming even if it's stopped.

davidt99:
I was trying to do it with a middleware but the declare queue callback doesn't get the actor option, as described in #91

ryanhiebert:
> I was trying to do it with a middleware but the declare queue callback doesn't get the actor option, as described in #91

Interesting. My thinking is that since queues may be shared between actors, pulling the max_priority off the actor probably doesn't make sense. And even then, the queue need not be declared on the actor, so it's at best an unreliable way to set it. I think what you've written here makes the most sense.

davidt99:
@Bogdanp I rebased and the tests are passing, can you take a look? There are more improvements that I want to make into dramatiq but this PR is on the shelf for almost a month...

Bogdanp:
Thanks! I merged most of your changes with rebase.  I added explicit support for pausing consumers in tests as opposed to what you had w/ not starting them at all.

PR #167, opened 2019-02-09
Travis updates

Bogdanp:
This is blocked by Travis not supporting RMQ on Xenial. We'll either have to do our own thing or move to a different CI system (like Semaphore).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,18 @@
 language: python
 sudo: required
+dist: xenial
 python:
   - "3.5"
   - "3.6"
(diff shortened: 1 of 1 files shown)


rpkilby:
Commenting out of curiosity. It looks like RMQ doesn't have first class support, but there is apparently a [workaround](https://travis-ci.community/t/rabbitmq-on-xenial/1827). Does this work? If not, any particular reason?

Bogdanp:
Sweet! I wasn't aware of the workaround. Thanks, Ryan!

PR #169, opened 2019-02-12
Allow to specify an actor class in actor decorator.

### LINKED ISSUE #163: Allow passing a custom actor class in @actor decorator
Hi,

Many thanks for working on this project. It's great for those of us who think bloated software is no longer an option (image).
 

# Issues

## Short description

It would really help should the `@actor` decorator took an `actor_class` argument which would default to `Actor`. This argument would define the class that will be used to instantiate the actors.

## Long description

We are instantiating our workers in a Flask CLI and rely on the [app factories](http://flask.pocoo.org/docs/1.0/patterns/appfactories/) pattern. So when all the imports happen (and therefore when the `@actor` decorator is applied on actors functions), we still have no idea on what the broker should be. We can only instantiate the broker later on, once the config is loaded and the Flask app created.

Then, the `Actor.__init__` function seems to produce some side effects on the default broker (the actor is somewhat coupled with a broker, which I don't find great). So once the `@actor` decorator is applied, it's already too late unfortunately.

Our workaround to this was to create a custom `LazyActor` class which is declared as follow:

```python
class LazyActor(Actor):
  def __init__(self, *args, **kwargs):
    kwargs.pop('broker')
    self.__args = args
    self.__kwargs = kwargs

  def init_actor(self, broker):
    super().__init__(self, *self.__args, broker=broker, **self.__kwargs)
```

And then calling the `init_actor` function with the broker for each actor in our app factory:

```python
[…]
broker = RabbitmqBroker(…)
for lazy_actor in [my_actor_1, my_actor_2, ]:
  lazy_actor.init_actor(broker=broker)
[…]
```

But this prevent us from using the `@actor` decorator since objects instantiated by this decorator are forcefully of `Actor` type. So we had to duplicate the decorator in our code base, just changing `Actor` to `LazyActor` which is a bit sad.

If you agree with this tiny change, I could submit a PR. Or maybe you can think of something else that would fix our issue in a better, more elegant way?

From a more general perspective, I think that dramatiq makes slightly too much assumptions at import time. Maybe the `Actor` class should do things more lazily, once we're sure the user doesn't want to set things up by himself. This is one of the many things I despised in Celery.

## Checklist

* [x] What OS are you using?
* [x] What did you do?
* [x] What happened?

## What OS are you using?

Arch, but not relevant.

Latest, but not relevant.

## What did you do?

See Long Description.

Not relevant.

## What happened?

Not relevant.

gilbsgilbs:
Sorry, I had to move the decorator function down so that I can get the `Actor` class declared soon enough, which kind of messes up the diff.

Fixes #163

### DIFF
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -26,74 +26,6 @@ from .message import Message
 _queue_name_re = re.compile(r"[a-zA-Z_][a-zA-Z0-9._-]*")
 
 
-def actor(fn=None, *, actor_name=None, queue_name="default", priority=0, broker=None, **options):
-    """Declare an actor.
-
-    Examples:
-
-      >>> import dramatiq
-
-      >>> @dramatiq.actor
-      ... def add(x, y):
-      ...   print(x + y)
-      ...
-      >>> add
-      Actor(<function add at 0x106c6d488>, queue_name='default', actor_name='add')
-
-      >>> add(1, 2)
-      3
-
-      >>> add.send(1, 2)
-      Message(
-        queue_name='default',
-        actor_name='add',
-        args=(1, 2), kwargs={}, options={},
-        message_id='e0d27b45-7900-41da-bb97-553b8a081206',
-        message_timestamp=1497862448685)
-
-    Parameters:
-      fn(callable): The function to wrap.
-      actor_name(str): The name of the actor.
-      queue_name(str): The name of the queue to use.
-      priority(int): The actor's global priority.  If two tasks have
-        been pulled on a worker concurrently and one has a higher
-        priority than the other then it will be processed first.
-        Lower numbers represent higher priorities.
-      broker(Broker): The broker to use with this actor.
-      **options(dict): Arbitrary options that vary with the set of
-        middleware that you use.  See ``get_broker().actor_options``.
-
-    Returns:
-      Actor: The decorated function.
-    """
-    def decorator(fn):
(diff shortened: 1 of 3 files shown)


davidt99:
I would like to see the LazyActor you mentioned in #163 in this PR so it becomes "official" because I ran into a similar issue.

dramatiq/actor.py
```diff
@@ -220,3 +152,74 @@ def __repr__(self):
 
     def __str__(self):
         return "Actor(%(actor_name)s)" % vars(self)
+
+
+def actor(fn=None, *, actor_name=None, queue_name="default", priority=0, broker=None, actor_class=Actor, **options):
```
Bogdanp: Tiny nitpick: please move `actor_class` before `actor_name`. All of those are keyword-only arguments so it's not a breaking change and I like how it reads a little better.

Bogdanp:
@davidt99 I'd rather we mention this somewhere in the guide rather than include the code for `LazyActor`.

@gilbsgilbs I just have that one tiny little nitpick. Otherwise, this looks good to me.

dramatiq/actor.py
gilbsgilbs: Done.

gilbsgilbs:
Thanks for the review @Bogdanp .

Bogdanp:
@gilbsgilbs thanks again. This is now merged with rebase.

PR #182, opened 2019-03-05
feat(actors): add check for delay and max_age

virusdefender:


.gitignore
```diff
@@ -9,3 +9,4 @@
 /*.egg-info
 /htmlcov
 __pycache__
+.idea/
```
Bogdanp: You should add the `.idea` folder to your [local git excludes instead](https://help.github.com/en/articles/ignoring-files#explicit-repository-excludes).

dramatiq/actor.py
```diff
@@ -119,13 +119,17 @@ def send_with_options(self, *, args=None, kwargs=None, delay=None, **options):
...
             broker and any registered middleware.
 
         Returns:
           Message: The enqueued message.
         """
+        max_age = self.options.get('max_age', 0)
+        if delay and max_age and delay > max_age:
+            raise ValueError('delay should not greater than max_age')
+
         message = self.message_with_options(args=args, kwargs=kwargs, **options)
```
Bogdanp: This is not really the appropriate place for this check. Instead, you should add `delay` to the `options` dictionary and move the check into the `before_enqueue` method of the `AgeLimit` middleware.

PR #186, opened 2019-03-12
Add possibility to use a custom worker class

xelhark:
Added the possibility to use a custom worker class.

The typical usage would be to use a config file like this:

```
import dramatiq
from dramatiq.brokers.stub import StubBroker

broker = StubBroker()
dramatiq.set_broker(broker)

# Import tasks
from my.module.dramatiq_tasks import *

# Define your custom worker
from dramatiq.worker import Worker
class MyCustomDramatiQWorker(Worker):
    def __init__(self, *args, **kwargs):
        super(MyCustomDramatiQWorker, self).__init__(*args, **kwargs)
        do_more_init_stuff()

WORKER_CLASS = MyCustomDramatiQWorker
```

When starting a new worker, dramatiq will check whether there exists an attribute called `WORKER_CLASS` defined in the configuration module and it will instantiate workers with that class if possible.

If the class is not a subclass of the original Worker it will launch a RuntimeError informing the user that the worker class must inherit dramatiq.worker.Worker.

### DIFF
--- a/dramatiq/cli.py
+++ b/dramatiq/cli.py
@@ -293,7 +293,14 @@ def worker_process(args, worker_id, logging_pipe):
         for module in args.modules:
             importlib.import_module(module)
 
-        worker = Worker(broker, queues=args.queues, worker_threads=args.threads)
+        worker_class = Worker
+        if hasattr(module, 'WORKER_CLASS'):
+            worker_class = module.WORKER_CLASS
+            logger.info("Detected WORKER_CLASS %s" % worker_class)
+            if type(worker_class) is not type or not issubclass(worker_class, Worker):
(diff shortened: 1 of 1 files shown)


xelhark:
@Bogdanp this is the PR I mentioned in the other issue, to allow users to use a custom worker class. Let me know what you think, I'd love to use a released version, but of course it's your call here 👍

Bogdanp:
Thanks for your work on this @xelhark, but I won't be merging it since I want so avoid adding too many knobs to Dramatiq. Folks aren't supposed to depend on the `Worker` implementation either since I want to be able to make potentially interface-breaking changes to the worker in a way that is transparent to users.

PR #188, opened 2019-03-15
feat(middleware): Increase the management of middleware for handle_message

fanjindong:
Increase the management of middleware for _ConsumerThread.handle_message


Bogdanp:
Changes like these need to be discussed first. Please start a topic on Reddit and let us know why you need this.

PR #191, opened 2019-03-18
Access broker in object attributes

bersace:
Hi @Bogdanp,

I expected that Dramatiq would be able to resolve deep attribute access like you can define in `pkg_resources` entry-point. This PR add this behaviour, based on pkg_resources implementation ( https://github.com/pypa/setuptools/blob/ad98fcb350fa3ce63389520a391e23274de11064/pkg_resources/__init__.py#L2426-L2434 ).

This ease integration with e.g. Flask by accepting `dramatiq some_module:app.broker` construction.

What do you think of this ?

### DIFF
--- a/dramatiq/cli.py
+++ b/dramatiq/cli.py
@@ -20,6 +20,7 @@
 
 import argparse
 import atexit
+import functools
 import importlib
 import logging
 import multiprocessing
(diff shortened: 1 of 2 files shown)


Bogdanp:
This is good, but it needs a test before I can merge it. Check out `test_cli.py`.

bersace:
Hi @Bogdanp. Thanks for pointing at it. I choosed to update existing test to both check deep attribute accessing and instance check.

Appveyor doesn't have build logs for me. Do i need privileges or something else to have logs ?

bersace:
@Bogdanp beware it's fixup. You either need to squash commits or let me rebase --autosquash before merging ! Thanks !

Bogdanp:
Thanks. This has now been merged w/ rebase.

PR #194, opened 2019-03-20
fix pipe_ignore has effect on sending message

metheoryt:
fix issue when pipe_ignore has effect on sending message instead if receiving, as it described in documentation https://dramatiq.io/cookbook.html#pipelines


Bogdanp:
Thanks! Please add a test for this and don't forget to add yourself to `CONTRIBUTORS.md`.

metheoryt:
I need some clarification about actor whose messages gets piped - should he be declared with `store_result=True` to be able to receive message result as argument or this feature is enabled as long as dramatiq has result backend?

Bogdanp:
Thanks again! This is merged with rebase and will be released soon.

PR #198, opened 2019-03-31
Fix watchdog_gevent import.

douglasmiranda:
The late importing of watchdog_gevent will allow us to use, for example,  the polling watcher without having to install watchdog_gevent.

This discussion started here: https://www.reddit.com/r/dramatiq/comments/b72bdo/auto_reload_without_watchdog_gevent/


Bogdanp:
I made some changes on top of this and merged it with rebase. We'll fall back to `Observer` if `watchdog-gevent` is not available now.

PR #210, opened 2019-04-18
Multiple hosts for RabbitMQ

srecnig:
One can now pass a `parameters` argument to a `RabbitmqBroker` containing a list of connection parameters.

Open questions: 
* Should something like this be possible with URLs, like celery does it, e.g.: `amqps://foo:5672;amqps://bar:5672`? 
* Did i miss anything expect the obvious chores already documented in the contributing document (which i will do once everything is ready) – e.g. should i add this to the docs?
* this test: `test_rabbitmq_actors_can_delay_messages_independent_of_each_other` failed, but to me, it doesn't look like its connected to my change – is this a flaky test, or am i missing something here?


Bogdanp:
Thanks! I'll take a look at this on Sunday. Please do add this to the advanced section of the docs as well.

Bogdanp:
Thanks! I've merged this with rebase and it'll go out with 1.6.0.

srecnig:
thanks 👍 looking forward to using this!

PR #219, opened 2019-07-22
Update allowed prometheus-client version range

robinro:
The current version is 0.7.1 and works fine with dramatiq as far as I can see.


Bogdanp:
Thank! I've merged your change with rebase and will release version 1.6.1 later today.

PR #221, opened 2019-08-03
Update uncaught worker exceptions' level as error

### LINKED ISSUE #220: Shouldn't log severity be ERROR instead of WARNING when a problem occurs on worker?
I am using sentry to collect logs with `ERROR` severity. I think I should be handling `WARNING` ones as well, however, I had to go with `>=ERROR` since some third party libraries cause much unexpected noise with warnings.

Here are the lines:

https://github.com/Bogdanp/dramatiq/blob/d7da3c7034752e154e2a1fd917f8baab5643553b/dramatiq/worker.py#L485

https://github.com/Bogdanp/dramatiq/blob/d7da3c7034752e154e2a1fd917f8baab5643553b/dramatiq/worker.py#L487

These two lines run when message processing fails on the worker. And I feel like `WARNING` suits better for the cases when a task is done but something strange happens.

What do you think?

th0th:
Closes #220.

### DIFF
--- a/dramatiq/worker.py
+++ b/dramatiq/worker.py
@@ -482,9 +482,9 @@ class _WorkerThread(Thread):
             message.stuff_exception(e)
 
             if isinstance(e, RateLimitExceeded):
-                self.logger.warning("Rate limit exceeded in message %s: %s.", message, e)
+                self.logger.error("Rate limit exceeded in message %s: %s.", message, e)
             else:
-                self.logger.warning("Failed to process message %s with unhandled exception.", message, exc_info=True)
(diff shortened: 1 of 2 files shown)


th0th:
Is there anything I can do about the failing check?

Bogdanp:
You'll have to fix the `test_workers_log_rate_limit_exceeded_errors_differently` test since that's failing now.

Bogdanp:
Thanks and sorry about the delay. This is now merged into master and I'll cut a release in the next couple of weeks.

th0th:
Thanks for merging :)

Issue #227, opened 2019-09-16
Can't Catch Error in Dramatiq worker and crashing the MainThread


jrusso1020:
# Issues

GitHub issues are for bugs.  If you have questions, please ask them on the [discussion board](https://reddit.com/r/dramatiq).

## Checklist

* [x] Does your title concisely summarize the problem?
* [x] Did you include a minimal, reproducible example?
* [x] What OS are you using?
* [x] What version of Dramatiq are you using?
* [x] What did you do?
* [x] What did you expect would happen?
* [x] What happened?

## What OS are you using?

macOS 10.14.5

## What version of Dramatiq are you using?

1.6.1

## What did you do?

While running a dramatiq actor that involves querying an elastic search instance on AWS a critical error occurs causing the dramatiq MainThread to crash. I try to catch the error in the worker as well as around the piece of code I believe is crashing but I am unable to catch the exception and therefore unable to tell what exactly is happening. I have checked that the environment variables are set correctly.

The code works when I perform a local query to elasticsearch on my computer, but it fails when trying to query a Elasticsearch cluster on AWS. However, I am able to run the code in the python terminal. It just fails in the dramatiq worker

```
@dramatiq.actor
def worker():
    import os
    from elasticsearch_dsl import Search
    from elasticsearch import Elasticsearch, RequestsHttpConnection
    from requests_aws4auth import AWS4Auth
    index = os.getenv('ELASTIC_SEARCH_INDEX')
    host = os.getenv('ELASTIC_SEARCH_HOST')
    awsauth = AWS4Auth(os.getenv('AWS_ACCESS_KEY'), os.getenv('AWS_SECRET_KEY'), os.getenv('AWS_REGION'), 'es')
    es = Elasticsearch(
        hosts=[{'host': host, 'port': 443}],
        http_auth=awsauth,
        use_ssl=True,
        verify_certs=True,
        connection_class=RequestsHttpConnection
    )
    logger.info('got here')
    try:
        logger.info("es.info {}".format(es.info()))
    except Exception as err:
        logger.info(err)
```

## What did you expect would happen?

I would expect to be able to catch the exception or at least have the exception surfaced by dramatiq so that I can see what is happening and debug it. If I was able to see what exception/error was occuring that would help me know what is going wrong when trying to access the AWS elasticsearch cluster

## What happened?

Received the following critical logging message, and then the MainThread is shut down killing all dramatiq workers
```
[...]

PR #228, opened 2019-09-17
Add Spawn Flag

jrusso1020:
Add a `--spawn` flag to the cli that will allow a user to spawn
processes instead of fork processes if they are on an OS that
defaults to forking processes such as unix based machines.

This started from a conversation on this issue https://github.com/Bogdanp/dramatiq/issues/227
which showed the problem on unix(namely macOS) systems caused by
forking processes.

### DIFF
--- a/dramatiq/cli.py
+++ b/dramatiq/cli.py
@@ -163,6 +163,10 @@ def make_argument_parser():
         "--log-file", type=str,
         help="write all logs to a file (default: sys.stderr)",
     )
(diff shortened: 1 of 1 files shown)


Bogdanp:
That was quick! I'll play around with this tomorrow morning and then merge it.

Bogdanp:
I've merged the change w/ rebase and renamed the flag to `--use-spawn`. Thanks!

PR #230, opened 2019-10-13
Add forks

Bogdanp:


dramatiq/forks_set.py
```diff
@@ -0,0 +1,62 @@
...
+import time
+from ctypes import Array, Structure, c_bool, c_byte, c_int
+
+
+class Buffer(Array):
+    _length_ = 1024 * 1024
+    _type_ = c_byte
+
+
+class ForksSet(Structure):
```
Bogdanp: https://www.youtube.com/watch?v=4PLvdmifDSk

Bogdanp:
If anyone has a better name for this functionality other than `forks`/`fork functions`, I'm all ears. My imagination is failing me at the moment.

Bogdanp:
Also, if someone with access to a windows machine could give this a try I'd be super grateful.

Steps to try on windows:

1. clone the repo and checkout this branch
2. run `pip install '.[rabbitmq]'` in the root
3. `cd examples/basic`
4. `dramatiq example`
5. visit http://127.0.0.1:9191 and make sure the Prometheus exposition server is running

bersace:
What do you think of `side processes` ?

CaselIT:
Would you advise using a fork function when using apscheduler to for periodic jobs?
I have not found suggestions in the documentation regarding this, so I was wondering if it was advisable.



> 
> 
> Also, if someone with access to a windows machine could give this a try I'd be super grateful.
> 
> Steps to try on windows:
> 
>     1. clone the repo and checkout this branch
> 
>     2. run `pip install '.[rabbitmq]'` in the root
> 
>     3. `cd examples/basic`
> 
>     4. `dramatiq example`
> 
>     5. visit http://127.0.0.1:9191 and make sure the Prometheus exposition server is running

I bit late, but for what is worth, it does work on windows

bersace:
@CaselIT I think not. I'm maintaining periodiq, and periodiq scheduler must be a single process. Side processes purpose is for all your worker processes (e.g. prometheus exporter).

CaselIT:
Well, a fork function creates a new process for each one, so technically it is a single process as far as the scheduler is concerned.

Of course you cannot control it when starting it that was, so for a scheduler that has a cli it maybe not the best way, but `apscheduler` does not offer one.

Bogdanp:
@CaselIT the functionality is intended for side-loading code next to workers, so in that sense I think it's appropriate. You just need to be careful if you run multiple sets of workers.

CaselIT:
Thanks for the reply. Yes, of course in case of multiple workers only one of them should start the scheduler or multiple jobs could be triggered.

PR #236, opened 2019-11-01
MaxTasksPerChild middleware

Sovetnikov:
In our project, it should be possible to restart workers periodically to free memory after tasks with heavy memory usage.
I did working prototype of MaxTasksPerChild middleware that acts like celery max tasks per child parameter.
If there is interest in such feature in dramatiq i can fix issues if there comments about implementation and write unit test for it.
Will test this code in production environment in next week.

### DIFF
--- a/dramatiq/broker.py
+++ b/dramatiq/broker.py
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
+from dramatiq.middleware.middleware import RestartWorker
 from .errors import ActorNotFound
 from .logging import get_logger
 from .middleware import MiddlewareError, default_middleware
@@ -93,11 +93,16 @@ class Broker:
                 self.logger.critical("Unexpected failure in before_%s.", signal, exc_info=True)
 
     def emit_after(self, signal, *args, **kwargs):
+        restart_requested_by_middleware = False
         for middleware in reversed(self.middleware):
             try:
                 getattr(middleware, "after_" + signal)(self, *args, **kwargs)
+            except RestartWorker as e:
+                restart_requested_by_middleware = e
             except Exception:
                 self.logger.critical("Unexpected failure in after_%s.", signal, exc_info=True)
+        if restart_requested_by_middleware:
+            raise restart_requested_by_middleware
 
     def add_middleware(self, middleware, *, before=None, after=None):
         """Add a middleware object to this broker.  The middleware is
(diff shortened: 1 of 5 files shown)


Bogdanp:
Sorry, but I don't think I want to encourage this type of use. If you have code whose memory leaks can't be fixed for whatever reason (including good ones, like it doesn't make sense for the company to spend time on such things at the moment), I'd recommend having a shared process pool between the workers specifically for running that type of code and killing the child processes after use.

Sovetnikov:
There is no memory leak, processes can't free memory they using to the operating system, just restart ...

magraeber:
I have recently switched a project from Celery to Dramatiq. And the problem described here is exactly what happens to me. I'm running workers in production at small AWS EC2 instances with limited memory. 
With Celery I used this settings:
CELERY_MAX_TASKS_PER_CHILD 
CELERY_WORKER_MAX_MEMORY_PER_CHILD 
It worked very reliable. Now with Dramatiq I'm missing a possibility to periodically restart workers. Everything else works perfectly.

@Bogdanp Is there a chance to add a feature for it?
If not, can you provide some code snippets for you suggested solution?

That is a real life problem, that people have.
For example I have found this discussion:
https://www.reddit.com/r/dramatiq/comments/esoeek/process_task_then_restart/

Sovetnikov:
@magraeber you can check my fork of dramatiq with such feature https://github.com/Sovetnikov/dramatiq
I am using this code in a production, but there is one bug with message dequeuing when redis queue is empty that causes unnecessary worker restarts. This error does not affect the operations if there is no need for high frequency tasks, so haven't fixed it yet

magraeber:
Thanks @Sovetnikov 
I will have a look. BTW, I'm already using your django_periodiq package.

As motivation for putting such features directly in dramatiq. In [this](https://sudonull.com/post/6810-50-shades-of-celery) nice article I have found the following paragraph, which expresses exactly what I feel:

> Unfortunately, memory leaks arise very easily, and it is difficult to find and fix them.
>
> In general, working with memory in Python is very controversial. You will spend a lot of time and nerves to understand why a leak occurs, and then it turns out that it is not even in your code. Therefore, always starting a project, put a memory limit on the worker : worker_max_memory_per_child.

Bogdanp:
@magraeber here's a full example of how you might do this: https://github.com/Bogdanp/dramatiq/blob/9161d66bba6697043bcc0a439af50980820e5e96/examples/max_tasks_per_child/app.py . This is somewhat limited on Windows, where you need a process manager to restart the process when it stops, but you should have something like that in place for production anyway.

Sovetnikov:
@magraeber about error that i mentioned in my comment, just investigated it and the cause was bad data (empty messages) in my queues. Seems i got that when debugging my fork ... just cleanup my queues and all ok now

Issue #51, opened 2018-04-04
Feature request: aborting messages


rpkilby:
Would it be possible to abort messages by ID? Similar to celery's [task revocation](http://docs.celeryproject.org/en/latest/userguide/workers.html?highlight=revoke#revoke-revoking-tasks). Optionally, the [feature matrix](https://dramatiq.io/motivation.html#compared-to) could be updated to reflect that task cancellation isn't supported.

rpkilby:
Whoops, duplicate of #37.

Bogdanp:
Reopening this because enough people have requested it thus far that I'm likely to add it.

virusdefender:
I think it's an important feature, waiting~

b1ngz:
Revoking tasks is useful, any plan to add it?

Bogdanp:
I'm closing this since [dramatiq-abort](https://flared.github.io/dramatiq-abort/) implements this functionality.

ToGoBananas:
@Bogdanp Hi, any plans to implement cancelling for dramatiq-gevent?

I am getting [CRITICAL] Failed to set exception (Abort) in thread

olegstepura:
> I'm closing this since dramatiq-abort implements this functionality.

`dramatiq-abort` only allows to cancel tasks if using redis backend. RabbitMQ case seem to be not covered still.

PR #239, opened 2019-12-03
feat(abortable): support task abort

isra17:
This address #51 and #37 .

@Bogdanp, let me know if the design / solution look good to you and I would be willing to work on documentation and tests once the code part is accepted.

In short, each worker has a thread that monitor running tasks and listen to a keyed-event systems. A third-party can abort a task by adding a message id to the keyed-event systems, letting the monitor thread abort the running task using the same method as time-limit middleware or skip a dequeued task.

### DIFF
--- a/dramatiq/abortable/backend.py
+++ b/dramatiq/abortable/backend.py
@@ -1,0 +1,12 @@
+class EventBackend:
+    """ABC for event backends.
+    """
+
+    def wait_many(self, keys, timeout):  # pragma: no cover
+        raise NotImplementedError
+
+    def poll(self, key):  # pragma: no cover
+        raise NotImplementedError
+
+    def notify(self, key, ttl):  # pragma: no cover
+        raise NotImplementedError
--- a/dramatiq/abortable/__init__.py
+++ b/dramatiq/abortable/__init__.py
@@ -1,0 +1,4 @@
+from .backend import EventBackend
+from .middleware import Abort, Abortable
+
+__all__ = ["EventBackend", "Abortable", "Abort"]
--- a/dramatiq/abortable/backends/__init__.py
+++ b/dramatiq/abortable/backends/__init__.py
@@ -1,0 +1,14 @@
+import warnings
+
+from .stub import StubBackend
+
+try:
+    from .redis import RedisBackend
+except ImportError:  # pragma: no cover
+    warnings.warn(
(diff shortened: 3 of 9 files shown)


dramatiq/abortable/backend.py
```diff
@@ -0,0 +1,12 @@
+class EventBackend:
+    """ABC for event backends.
```
aviau: Might as well use the [abc lib](https://docs.python.org/3/library/abc.html)?

isra17: Rest of the projet don't, I prefer to stay consistent.

sauravmahuri2007: @isra17 ,

Finally the most awaited feature seems to be available soon.
**Can you please give an example for your code how to use the abort functionality to terminate a long running task (message)?**

isra17:
@sauravmahuri2007 you can check the test case, In short you need to add the `Abortable` middleware and then you are able to call `message.abort()` or `abortable_middleware.abort(message_id)`

Bogdanp:
I think this is great, but my preference would be for it to live in a separate package. While the code looks good and the approach seems sensible, I would rather not increase my maintenance burden at the moment. It looks like this doesn't need any cooperation from dramatiq itself (feel free to correct me if I'm wrong) apart from the `Message.abort` method, which could be a plain function:

```python
from dramatiq_abortable import abort

message = some_task.send()
abort(message)
```

So it should be easy to create a separate package for this functionality. I'd gladly add a cookbook section pointing at the package if you do decide to publish it.

isra17:
Sounds reasonnable, I will move it to its own package and create a PR for the cookbook.

sauravmahuri2007:
> @sauravmahuri2007 you can check the test case, In short you need to add the Abortable middleware and then you are able to call message.abort() or abortable_middleware.abort(message_id)

@isra17 ,

I did try aborting the execution using the message instance though didn't try using the message_id (as you mentioned in your second way). I was able to call message.abort() but it looked like the exception was only raised when the task was actually completed. It may be my mistake so I want to send you my sample test code. 
I was testing in Windows10 having Python3.8.0 with Redis-Server. Also it would be nice to have the abort functionality using the message id as most of the time we may not have the message instance or don't know how to get back the message instance using the message id.

cc: @Bogdanp

isra17:
Installable package is available here with docs: https://flared.github.io/dramatiq-abort/

PR #241, opened 2019-12-03
add `receive_in_kwargs` flag

evstratbg:
When pipelined, It may be useful to pass results of a previous actor to another one within kwargs, with a actor_name as a key. Thus, we can select relative data for sure


Bogdanp:
Sorry, but I don't see the utility in this.

PR #242, opened 2019-12-06
beautify imports

evstratbg:


gdvalle:
Wouldn't this break for users who don't have extras installed for each broker?

Bogdanp:
@gdvalle is right, sorry @evstratbg!

Issue #19, opened 2017-12-18
[feature request] add crontab like interface to schedule tasks like in tasktiger


cailiang9:
related interfaces can be seen here: https://github.com/closeio/tasktiger

jssuzanne:
+1 cron service could be a good feature

Bogdanp:
I've thought about this and ultimately decided that implementing something like this would mean building in a lot of functionality that already exists in things like [`apscheduler`](https://apscheduler.readthedocs.io/en/latest/) so I'm just going to recommend that people use that in conjunction with Dramatiq.

I'll probably write a little how-to on using the two of 'em together and link it here when I do.

rakanalh:
APScheduler example:
```
import os
import django

os.environ.setdefault("DJANGO_SETTINGS_MODULE", "settings")
django.setup()

from apscheduler.schedulers.background import BlockingScheduler  # noqa

from myapp.tasks import MyTask  # noqa


if __name__ == '__main__':
    scheduler = BlockingScheduler()
    scheduler.add_job(MyTask.send, 'interval', minutes=1)
    scheduler.start()
    print('Press Ctrl+{0} to exit'.format('Break' if os.name == 'nt' else 'C'))
    scheduler.shutdown()

```

ericgj:
Hi, ran into problems with using APScheduler in this way.  It pickles arguments, so any `Actor.send` is going to need `self` (i.e. the actor) as the first argument, and actors can't be pickled.  I tried various things but it seems impossible to use bound methods as jobs -- with or without additional arguments. Really it's an issue for APScheduler I suppose but have others found a way around this in the context of Dramatiq?

Bogdanp:
Hi @ericgj! If you can't use the `MemoryJobStore` (which doesn't use pickle), I'd write dispatcher functions that *can* be pickled. For example:

```python
def dispatch_some_task(*args, **kwargs):
  return some_task.send(*args, **kwargs)

if __name__ == "__main__":
  scheduler = BlockingScheduler()
  scheduler.add_jobstore("redis")
  scheduler.add_job(dispatch_some_task, args=(1, 2,), 'interval', minutes=1)
  scheduler.start()
```

If you want to get a little crazy you could do something like this:

`cron.py`:
```python
import importlib 

from apscheduler.schedulers.background import BackgroundScheduler
from apscheduler.triggers.cron import CronTrigger
from functools import wraps

scheduler = BackgroundScheduler()
[...]

PR #248, opened 2019-12-15
doc: motivation for not implementing cronlike scheduling

while0pass:
#19

### DIFF
--- a/docs/source/motivation.rst
+++ b/docs/source/motivation.rst
@@ -54,7 +54,7 @@ of the main differences between Dramatiq, Celery and RQ:
 |                              |          |             |              |              |
 +------------------------------+----------+-------------+--------------+--------------+
 | Cronlike scheduling          | No       | Yes         | Yes          | No           |
-|                              |          |             |              |              |
+|                              | [#cron]_ |             |              |              |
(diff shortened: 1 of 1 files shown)


Bogdanp:
Thanks! I've made some edits and merged this with rebase.

jcass77:
I have been using Dramatiq + APScheduler to schedule tasks and, although it works well, there is a gotcha here that I have seen trip people up: using a `BackgroundScheduler` and a persistent job store can cause multiple instances of the scheduler to run concurrently (since schedulers can be started up in different Django web server processes). People might try doing something like this in an attempt to dynamically schedule jobs from within their Django application, for example.

The recommended approach can **only** be used safely using a `BlockingScheduler` in a custom Django management command (and by implication, only for fixed execution schedules as jobs cannot be added once the scheduler has been started). This limitation is caused by the fact that APScheduler does not currently offer an [inter-process synchronization and signalling scheme](https://apscheduler.readthedocs.io/en/latest/faq.html#how-do-i-share-a-single-job-store-among-one-or-more-worker-processes) which would be required to coordinate job scheduling and execution across multiple schedulers running concurrently.

It might be worth mentioning this explicitly in the documentation as well.

PR #259, opened 2020-01-07
Fix redis lua script errors 

### LINKED ISSUE #258: [Redis broker] Error "user_script:90: too many results to unpack" causes worker to sit idle forever
## Checklist

* [ ] Did you include a minimal, reproducible example?
* [x] What OS are you using?
* [x] What did you do?
* [x] What happened?

## What OS are you using?

Debian GNU/Linux 9.6 (stretch)

We are using the `python:3.7.1-stretch` Docker image, if that helps.

1.7.0

## What did you do?

No particular action caused this. This seems to happen on its own after a while of running dramatiq in production. We are running 16 dramatiq workers at a time with 1 process and 32 threads each.

I expected the dramatiq workers would process tasks indefinitely, or restart as the result of an error

## What happened?

The worker fails to fetch new tasks to run, but does not die, so it sits idle forever (or until I restart it myself). Over the course of a few days, this will eventually happen to enough workers that tasks start backing up in the queue and we experience problems visible to users of our application.

Here is the log line containing the stack trace:

```
[dramatiq] message="Consumer encountered an unexpected error.", exc_info="Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/dramatiq/brokers/redis.py", line 315, in __next__
    data = self.message_cache.pop(0)
IndexError: pop from empty list

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.7/site-packages/dramatiq/worker.py", line 253, in run
    for message in self.consumer:
  File "/usr/local/lib/python3.7/site-packages/dramatiq/brokers/redis.py", line 328, in __next__
    self.prefetch - self.message_refc,
  File "/usr/local/lib/python3.7/site-packages/dramatiq/brokers/redis.py", line 256, in do_dispatch
    return dispatch(args=args, keys=keys)
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 2694, in __call__
    return client.evalsha(self.sha, len(keys), *args)
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 1944, in evalsha
    return self.execute_command('EVALSHA', sha, numkeys, *keys_and_args)
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 573, in execute_command
    return self.parse_response(connection, command_name, **options)
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 585, in parse_response
    response = connection.read_response()
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 582, in read_response
    raise response
redis.exceptions.ResponseError: Error running script (call to f_605aad1746512fb86e8ee6fa6129dccdf32eed59): @user_script:90: user_script:90: too many results to unpack"
```

While it is possible that a different error causes the issue (or that multiple errors can cause the issue), I believe this is causing it because the exception raised is not a subclass of `redis.exceptions.ConnectionError`. Since only `ConnectionError`s trigger restarts in [`dramatiq.brokers.redis`](https://github.com/Bogdanp/dramatiq/blob/v1.7.0/dramatiq/brokers/redis.py#L341), I believe the worker never gets restarted.

I plan to assist in getting this fixed, but I figured I'd post this first to make sure I'm on the right track. I propose doing the following:

1. The function call on line 90 in [dispatch.lua](https://github.com/Bogdanp/dramatiq/blob/v1.7.0/dramatiq/brokers/redis/dispatch.lua#L90) should be fixed to push message ids in batches so we don't fill up lua's stack.
1. Restart the worker on any uncaught exception in `dramatiq.brokers.redis`.

benekastah:
Calling `unpack` on a list with `7999` items or more causes lua to throw an error (reason why is in the code comments). This will fix errors of that kind in the lua script (primarily encountered in the `do_maintenance` branch).

At least partially fixes #258 (maybe fixes the whole thing; time will tell).

### DIFF
--- a/dramatiq/brokers/redis/dispatch.lua
+++ b/dramatiq/brokers/redis/dispatch.lua
@@ -75,6 +75,42 @@ for i=8,#ARGV do
     ARGS[i - 7] = ARGV[i]
 end
 
+
+-- In my testing I found the max size that we can unpack to be 7999 (which
+-- is LUAI_MAXCSTACK - 1). Lua can be built with a different LUAI_MAXCSTACK
+-- value, but redis doesn't appear to support doing that at the moment.
+-- https://github.com/antirez/redis/blob/fc0c9c8097a5b2bc8728bec9cfee26817a702f09/deps/lua/src/luaconf.h#L438
+local MAX_UNPACK_SIZE = 7999
+
+-- Iterates over a table in chunks, yielding a new chunk on each
+-- iteration. Used to do work in batches so we don't overflow lua's stack.
+--
+-- Example:
+--
+--   for chunk in iter_chunks(some_table) do
+--       do_something(unpack(chunk))
+--   end
+local function iter_chunks(tbl, chunksize)
+    if chunksize == nil then
+        chunksize = MAX_UNPACK_SIZE
+    end
+    local len = #tbl
+    local i = 1
+    return function()
+        if i <= len then
+            local chunk = {}
+            local last_idx = math.min(i + chunksize - 1, len)
+            local j
+            for j = i, last_idx do
+                table.insert(chunk, tbl[j])
+            end
+            i = last_idx + 1
+            return chunk
+        end
+    end
+end
+
+
 -- Every call to dispatch has some % chance to trigger maintenance on
 -- a queue.  Maintenance moves any unacked messages belonging to dead
 -- workers back to their queues and deletes any expired messages from
@@ -87,7 +123,9 @@ if do_maintenance == "1" then
         local dead_worker_queue_acks = dead_worker_acks .. "." .. queue_name
         local message_ids = redis.call("smembers", dead_worker_queue_acks)
         if next(message_ids) then
-            redis.call("rpush", queue_full_name, unpack(message_ids))
+            for message_ids_batch in iter_chunks(message_ids) do
+                redis.call("rpush", queue_full_name, unpack(message_ids_batch))
+            end
             redis.call("del", dead_worker_queue_acks)
         end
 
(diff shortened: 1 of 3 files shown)


Bogdanp:
Thanks! I'll dig into this on Saturday.

benekastah:
@Bogdanp I may have misread the test failure output, but it looks like the redis tests are failing because it can't persist to disk. Maybe the disk is full?

dramatiq/brokers/redis/dispatch.lua
```diff
@@ -75,6 +75,42 @@ for i=8,#ARGV do
     ARGS[i - 7] = ARGV[i]
 end
 
+
+-- In my testing I found the max size that we can unpack to be 7999 (which
+-- is LUAI_MAXCSTACK - 1). Lua can be built with a different LUAI_MAXCSTACK
```
Bogdanp: Is there any way we can get this exact value at runtime?

Bogdanp:
> Maybe the disk is full?

Most likely, the process is not allowed to write to disk and the test changes that you've made push it past the number of commands (10k by default?) it can run before persisting. I'll see what I can do about it next weekend.

Have you started running these changes in production? If so, do they seem to have fixed the problems you ran into?

This looks good, but I didn't get a chance to dive in as deep as I would've liked this weekend so I'm going to hold off merging until the next weekend.

dramatiq/brokers/redis/dispatch.lua
benekastah: Unfortunately it's only exposed in lua's C api as far as I can tell. AFAICT, there isn't a way to get those from lua code.

benekastah:
I haven't run this in production yet. Instead we temporarily set the maintenance chance to 0 since it was consistently failing for us. We may try running this in production next week, as well as the other PR I put up (which I think has a better chance of fixing our "lazy worker" issue).

I did run the new tests I wrote on the old code and verified that it failed, so the fact that they pass for me convinces me it will get the maintenance code working again.

Bogdanp:
Thanks! I made a couple changes on top of this and fixed the CI and the changes are now in master. They'll be release with 1.8.

PR #262, opened 2020-01-09
[Redis broker] `message_refc` dips into negative numbers on `ConnectionError` 

benekastah:
This can cause prefetching too many messages. The reason is that when we
try to call `consumer.ack` or `consumer.nack`, and a `ConnectionError`
happens, we will retry. But each time we retry, we decrement `message_refc`
again. Solved this by using a set to track queued messages, so calling `ack`
or `nack` on the same message doesn't mess up our accounting.

### DIFF
--- a/dramatiq/brokers/redis.py
+++ b/dramatiq/brokers/redis.py
@@ -273,8 +273,12 @@ class _RedisConsumer(Consumer):
         self.timeout = timeout
 
         self.message_cache = []
-        self.message_refc = 0
+        self.queued_message_ids = set()
         self.misses = 0
+
+    @property
+    def message_refc(self):
+        return len(self.queued_message_ids) + len(self.message_cache)
 
     def ack(self, message):
         try:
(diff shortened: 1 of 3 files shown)


dramatiq/brokers/redis.py
```diff
@@ -273,9 +273,13 @@ def __init__(self, broker, queue_name, prefetch, timeout):
         self.timeout = timeout
 
         self.message_cache = []
-        self.message_refc = 0
+        self.queued_message_ids = set()
         self.misses = 0
 
+    @property
+    def message_refc(self):
```
benekastah: This property is so I can use the same test code with this version of the file and with the previous version of the file.

dramatiq/brokers/redis.py
```diff
@@ -316,12 +320,14 @@ def __next__(self):
...
                     message = Message.decode(data)
+                    self.queued_message_ids.add(message.message_id)
                     return MessageProxy(message)
                 except IndexError:
                     # If there are fewer messages currently being
                     # processed than we're allowed to prefetch,
                     # prefetch up to that number of messages.
                     messages = []
+                    num_messages = len(self.queued_message_ids)
                     if self.message_refc < self.prefetch:
```
Bogdanp: Seems like this is a bug since `message_refc` has been removed. Did you mean to use `num_messages` here?

Bogdanp:
I made a couple changes here and merged this into master with rebase. Thanks for the fix!

PR #263, opened 2020-01-12
Add a skip-logging flag to the CLI

whalesalad:
This PR adds a `skip-logging` flag to the CLI invocation arguments.

This came from wanting to adjust the LOGFORMAT of Dramatiq's log messages. I wanted to eliminate the `[Thread-X]` bit and modify the order of things to better match the rest of my system

`logging.basicConfig` is a no-op if the root logger has handlers added to it, but due to the fact that the Dramatiq CLI is invoked before any of my code, there is no way to beat that to the punch.

So instead I opted to add a `--skip-logging` flag that will prevent `basicConfig` from ever happening. This allows the end user to define the handlers and whatnot without having to undo anything done by Dramatiq's bootup sequence.

Hopefully this is something that is worth merging into the project.

### DIFF
--- a/dramatiq/cli.py
+++ b/dramatiq/cli.py
@@ -172,6 +172,11 @@ def make_argument_parser():
         help="write all logs to a file (default: sys.stderr)",
     )
     parser.add_argument(
+        "--skip-logging",
+        action="store_true",
+        help="do not call logging.basicConfig()"
+    )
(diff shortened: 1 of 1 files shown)


whalesalad:
It looks like the CI failure was unrelated to this change.

Bogdanp:
Thank you!

Issue #206, opened 2019-04-10
pipeline doesn't respect (at least) first message delay option


metheoryt:
if i create pipeline as below and try to run it:
```
pipe = actor.message_with_options(delay=60_000) | actor.message()
pipe.run()
```
the pipe will be executed immediately though.
I think, pipeline should respect message delays, so e.g. if both first message delay and pipe depay specified, they should summarize.

synweap15:
I think currently only the `pipe.run(delay=60_000)` invocation will take effect, the `delay` parameter of messages will be ignored:

1. For the first message in a pipe, the `run` method named argument is taken, the message's one isn't considered:
https://github.com/Bogdanp/dramatiq/blob/952667885c53684912cefa4825eafd613ae5d2cd/dramatiq/composition.py#L111

2. For next-in-line messages, there's no delay taken into consideration:
https://github.com/Bogdanp/dramatiq/blob/843bc18f184f542918b7bd2539c3620fa7b635a3/dramatiq/middleware/pipelines.py#L56

I've got test cases and a fix ready, however in contrast to what @metheoryt proposes, I think it would be best to follow this logic:

1) If there's either a delay specified for the first message or there's a delay specified in pipe run invocation, take the one that's present
2) If they are both present, take the one that's bigger (instead of adding them).

And of course, the messages after the first message in a pipe should have their delays respected as well.

The delay parameter is specified as:

PR #264, opened 2020-01-13
Fixing delays of messages in a pipeline

synweap15:
Fixes https://github.com/Bogdanp/dramatiq/issues/206.

The delays of messages in the pipeline were ignored, only the `delay` parameter of pipeline's `run` method was taken into consideration.

This PR fixes the handling of delay parameters of messages in a pipeline (first and all following). In addition in alters the behavior of the `delay` parameter of pipeline's `run` method, so that:

1. If there's either a delay specified for the first message or there's a delay specified in pipe run invocation, take the one that's present.
2. If they are both present, take the one that has higher value.


Bogdanp:
Thanks!

PR #268, opened 2020-01-21
[Redis] Fix key error in queued message tracking

benekastah:
I forgot that removing a key from a set that doesn't have that key raises a KeyError.


Bogdanp:
Thanks!

PR #277, opened 2020-02-26
add retries

evstratbg:
Here I added Retry error to catch it and requeue

### DIFF
--- a/dramatiq/errors.py
+++ b/dramatiq/errors.py
@@ -66,3 +66,12 @@ class ConnectionClosed(ConnectionError):
 class RateLimitExceeded(DramatiqError):
     """Raised when a rate limit has been exceeded.
     """
+
+
+class Retry(DramatiqError):
+    """Raised a message should be retries.
+    """
+
+    def __init__(self, message, delay=None):
+        super(DramatiqError).__init__(message)
+        self.delay = delay
--- a/dramatiq/worker.py
+++ b/dramatiq/worker.py
@@ -23,7 +23,7 @@ from queue import Empty, PriorityQueue
 from threading import Event, Thread
 
 from .common import current_millis, iter_queue, join_all, q_name
-from .errors import ActorNotFound, ConnectionError, RateLimitExceeded
+from .errors import ActorNotFound, ConnectionError, RateLimitExceeded, Retry
 from .logging import get_logger
 from .middleware import Middleware, SkipMessage
 
@@ -474,6 +474,10 @@ class _WorkerThread(Thread):
         except SkipMessage:
             self.logger.warning("Message %s was skipped.", message)
             self.broker.emit_after("skip_message", message)
+
+        except Retry as e:
+            self.logger.warning("Message %s is retrying with %s delay.", message, e.delay)
+            self.broker.enqueue(message=message, delay=e.delay)
 
         except BaseException as e:
             # Stuff the exception into the message [proxy] so that it


Bogdanp:
Thanks for this, but I'd prefer users built this for themselves if they need it.

takhs91:
Any seconds thought on that?
 I was planning to use that or something similar. There are cases where known exceptions happen in a task and the task should be better retried without reporting an error (either on caller or on worker), since its a handled condition and it would simply pollute the logs, sentry etc.

What I didn't like so much about this specific implementation is fact that it won't increase the retries count but thats an easy fix, simple call again `self.broker.emit_after("process_message", message, exception=e)` instead of `enque` so it will pass from the retries middleware and probably add the delay as parameter. I would probably also log this retry event as info and not as warning.

dimrozakis:
Hi @Bogdanp 

I see that you closed this PR and replaced it with #295 which adds support for silent retries.

From my perspective, the added value of this PR isn't just that the task code can signal a retry without logging an error (which #295 provides), but that the task code can define the retry delay, overriding the default retry policy.

I can think of valid use cases for this feature. For example, consider tasks that communicate with an external API. In general, one would want these tasks to be retried with an exponential retry policy if something went wrong. This would work well for the case of network issues, external service being down etc. But, if the API call returns a rate-limit response with the instruction to try again in 2 minutes, it'd make sense to catch that in the task code with a `try/except` statement, and raise a `Retry` error with a delay of 2 minutes (or something slightly longer than that).

So the overall exponential retry policy is good for arbitrary errors, but then there will be cases where the task will know what is causing the error and when that error will be overcome, so it would make sense to retry itself at that time.

>Thanks for this, but I'd prefer users built this for themselves if they need it.

Can you please clarify how users could build this for themselves? If I remember correctly, the `retry_when` function can only signal whether a retry should occur or not, while being unable to define the retry's delay. Am I missing something else here? Could you please provide a hint as to how a user could build this for themselves without using a forked & modified version of dramatiq?

Bogdanp:
> Can you please clarify how users could build this for themselves? If I remember correctly, the retry_when function can only signal whether a retry should occur or not, while being unable to define the retry's delay. Am I missing something else here? Could you please provide a hint as to how a user could build this for themselves without using a forked & modified version of dramatiq?

Here's some (untested) pseudocode that does roughly the same thing that this PR did without modifying dramatiq at all:

```python
import dramatiq

class Retry(Exception): 
  def __init__(self, delay=30_000):
    self.delay = delay

def actor(**options):
  def decorator(fn):
    def wrapper(*args, **kwargs)
      try:
        return fn(*args, **kwargs)
      except Retry as e:
        the_actor.send_with_options(args=args, kwargs=kwargs, delay=e.delay)
     the_actor = dramatiq.actor(wrapper)
     return the_actor
  return decorator

@actor()
def flaky():
  if time() % 2 == 0:
    raise Retry(15000)
```

takhs91:
I don't think this is equivalant. Every time the task fails it sends a new message. The original caller can't get the result back.

dimrozakis:
@Bogdanp thanks for providing an example. However I think that @takhs91 is correct. This doesn't actually "retry" the task as much as it marks the task as completed with no return value and then schedules a new identical task. So due to it being a new task, the caller won't be able to access its result.

For the above scenario to work properly, I'm under the understanding that the same amqp message would have to be re-queued, which is what I think this MR accomplishes. Without it, perhaps this could be accomplished by the user by means of a custom `retry_when`, but then dramatiq's definition of `retry_when` would need to be extended in order to not just return a boolean, but optionally an integer in milliseconds for a custom delay. I can't think of a way to implement custom retry delays on the user's side without modifications to dramatiq's code.

Bogdanp:
@dimrozakis Ah, my bad! I lost track of exactly what the original PR was doing this morning. 

You can grab the current message using the [CurrentMessage](https://dramatiq.io/reference.html#dramatiq.middleware.CurrentMessage) middleware so instead of sending a new message, you can re-enqueue that same message.

Roughly:

```python


def actor(**options):
  def decorator(fn):
    broker = dramatiq.get_broker()
    def wrapper(*args, **kwargs)
      try:
        return fn(*args, **kwargs)
      except Retry as e:
        message = CurrentMessage.get_current_message()
        broker.enqueue(message, delay=e.delay)
     return wrapper
  return decorator
```

dimrozakis:
@Bogdanp thanks for the updated example!

>You can grab the current message using the CurrentMessage middleware so instead of sending a new message, you can re-enqueue that same message.

This does solve the issue of *retrying* as opposed to *cloning* the task, nice!

>You are right, though, that this won't do exactly the right thing when storing results, but neither would this PR.

Just to make sure we're on the same page, are you referring to an empty result being stored from the task that raised `Retry`, until the retried task's result becomes available?

>I think it may make sense for us to add a delay argument to the Retry exception in dramatiq and that ought to provide a nice interface for what you're trying to do.

I'm glad to hear you're considering this. It would provide a very handy interface to do what would otherwise require several tricks implemented client side to properly handle both retries & results. And could prove useful in any scenario where a task has contextual information that it can use to decide the proper (or optimal) time of a retry.

Bogdanp:
> Just to make sure we're on the same page, are you referring to an empty result being stored from the task that raised Retry, until the retried task's result becomes available?

Yep.

I'll look at adding `delay` to `Retry` this weekend.

Issue #279, opened 2020-02-27
Feature Request: emit warning when returning value but not storing results


rsyring:
https://github.com/Bogdanp/dramatiq/issues/177 saved me from a long time troubleshooting. But I wonder if it would make sense to emit a warning when the actor returns a value but no results are being saved?

Bogdanp:
I'd happily accept a PR for this! I'd recommend logging a warning instead of using the `warnings` module. I assume that's what you meant anyway, but just wanted to make sure.

synweap15:
Interesting idea. I'd be willing to take on this. Any preference when it should appear - when the worker is started, or when job is ran?

Bogdanp:
I'd go with the simplest approach: when the job is run.

Bogdanp:
Added in #281.

CaselIT:
This causes unnecessary warnings if the `Results` middleware is instantiated with `store_results=True` and the actors do not also specify `store_results=True`

Maybe `broker.declare_actor` should set a property on the actor that indicates if the results are stored by default?

Bogdanp:
Thanks, @CaselIT. I'll address that problem.

LincolnPuzey:
@CaselIT I believe your issue was fixed in 637a14b665e0996858c0a55cb7612b53d2f46ab4, I will close this now.

CaselIT:
thanks!

PR #281, opened 2020-03-06
Emit warning when returning value but not storing results

synweap15:
This adresses #279. It's a simple solution that will show a warning every time an actor that returns not None value is called, and there's no `store_results=True` (key is missing or it's false) in the actor's configuration.

I was thinking about other options, we could eg. add another middleware that would be always attached and just show warning, but I decided that we can start with the simplest option and let's see what @Bogdanp thinks about it :)

### DIFF
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -142,7 +142,14 @@ class Actor:
         try:
             self.logger.debug("Received args=%r kwargs=%r.", args, kwargs)
             start = time.perf_counter()
-            return self.fn(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


Bogdanp:
Looks great. I'll merge it tomorrow morning.

Bogdanp:
This is now merged with rebase. Thanks again!

darklow:
This causes lot of unnecessary warnings as task itself can be also called directly (for tests, etc) and result can be used in various ways. Would be great if these particular warnings could be made quiet. Results can be still useful in many ways even if you don't want to store results in some backend.

PR #282, opened 2020-03-12
Add support for saving message failures

### LINKED ISSUE #156: Backend doesn't store message failures
When working on adding tests for #151, I noticed that the backend doesn't support storing messages failure (failed manually or unhandled exception), causing a timeout when trying to get the result.
I think we need at least store the indication of the message failure. Another approach is to try to raise the exception that was raised in the worker.

takhs91:
Fixes #156 

Hi again @Bogdanp ! I progressed on that. I finally took your proposed approach of "forking" the results package. I would like to hear your thoughts.

I marked it as WIP because I also want to add more information about the last exception that led to the failure

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -41,3 +41,4 @@ of those changes to CLEARTYPE SRL.
 | [@asavoy](https://github.com/asavoy)                  | Alvin Savoy            |
 | [@benekastah](https://github.com/benekastah)          | Paul Harper            |
 | [@timdrijvers](https://github.com/timdrijvers)        | Tim Drijvers           |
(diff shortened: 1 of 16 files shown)


takhs91:
I added information related to the exception that caused the failure in the exception raised by get_result after retrieving the failure. One of the checks failed, it seems to be a timing issue, maybe they it needs to be less strict

Bogdanp:
Thanks. I'll try to take a look at this sometime this weekend or the next.

Bogdanp:
Thanks again @takhs91! I ended up going a similar but different route in #293, which I think provides a way forward that's both forwards and backwards-compatible w/o much bloat.

PR #291, opened 2020-03-27
Expose pipe's fileno to StreamablePipe

takhs91:
If its not exposed any given tasks module that imports libraries that may internally call faulthandler.enable() will produce an exeption

```
AttributeError: 'StreamablePipe' object has no attribute 'fileno'
```


Bogdanp:
Thanks! I've merged this with rebase.

PR #293, opened 2020-03-28
results: store actor failures

### LINKED ISSUE #156: Backend doesn't store message failures

Bogdanp:
Closes #156.  This should be both forwards (mostly) and backwards-compatible.  The only weirdness here is the interaction of failures with retries, but I think I'm fine with leaving the burden on the user to figure that out.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -14,7 +14,9 @@ Added
 
 * A custom Redis connection can now be passed to the Redis broker via
   the new ``client`` keyword argument.  (`#274`_, `@davidt99`_)
+* Support for storing actor exceptions.  (`#156`_)
 
+.. _#156: https://github.com/Bogdanp/dramatiq/issues/156
 .. _#274: https://github.com/Bogdanp/dramatiq/issues/274
 .. _@davidt99: https://github.com/davidt99
 
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -196,3 +196,5 @@ Errors
    :members:
 .. autoclass:: dramatiq.results.ResultTimeout
    :members:
+.. autoclass:: dramatiq.results.ResultFailure
+   :members:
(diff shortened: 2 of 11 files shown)


takhs91:
Hi, I mostly prefer your approach but I also believe the interaction of failures with the retries do not help the general case.

I believe the result should be final. At least from the caller's side. Maybe there is a value in getting intermediate failures, but I believe the most used case is to send a task and either get a result when it succeeded or it failed for good (e.g passed max retries ) especially when you block for it.

What were your thoughts of my original approach, that sets the final exception as a result in the `after_nack` method https://github.com/Bogdanp/dramatiq/pull/282/files#diff-2d67f34e8f8c952ee41a899bfc788304

Bogdanp:
> I believe the result should be final. At least from the caller's side. Maybe there is a value in getting intermediate failures, but I believe the most used case is to send a task and either get a result when it succeeded or it failed for good (e.g passed max retries ) especially when you block for it.

I agree, but the reality is that if someone uses the default retry settings, then it could be a month before a "result" is available with that approach.  I think, given the backwards-compatibility constraint, we can't do much better than this and the user of the library will have to make sure to use retry settings that make sense for their use case.

> What were your thoughts of my original approach, that sets the final exception as a result in the after_nack method.

One problem with that approach is I don't think there's a guarantee that an exception will always be available at that point (nor do I think there should necessarily be one). But the bigger issue is it still may be a long time before a message is rejected w/ the default settings. 

Then again, maybe that's OK. It might make the most sense to fail results on nack and just deal w/ the case when there's no exception head-on. I'll sleep on it and probably make that change tomorrow morning unless I think of a reason not to.

PR #294, opened 2020-03-28
Set daemon flag to False for workers

takhs91:
Previous conversation in https://github.com/Bogdanp/dramatiq/pull/289
Opened new after change of scope


Bogdanp:
Thanks again! I rebased and merged your changes.

Issue #290, opened 2020-03-23
Change `broker_priority` by Middleware


thomazthz:
First of all, thanks for this awesome lib! :)

I'm trying to set the priority of RabbitMQ in a Middleware but I realized that the [`before_enqueue` ](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/brokers/rabbitmq.py#L306) callback is called after the [`pika.BasicProperties` creation](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/brokers/rabbitmq.py#L287).

Maybe these properties should be defined after the `before_enqueue` callback so we can change it with a Middleware. What do you think?

I don't know if this change violates the lib design or the code style. Please let me know if it make sense so I can open a PR.

### Possible approaches:

1.

```diff
diff --git a/dramatiq/brokers/rabbitmq.py b/dramatiq/brokers/rabbitmq.py
index af31b83..54aa4e4 100644
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -284,11 +284,6 @@ class RabbitmqBroker(Broker):
         queue_name = message.queue_name
         self.declare_queue(queue_name, ensure=True)
 
-        properties = pika.BasicProperties(
-            delivery_mode=2,
-            priority=message.options.get("broker_priority"),
-        )
-
         if delay is not None:
             queue_name = dq_name(queue_name)
             message_eta = current_millis() + delay
@@ -304,6 +299,10 @@ class RabbitmqBroker(Broker):
             try:
                 self.logger.debug("Enqueueing message %r on queue %r.", message.message_id, queue_name)
                 self.emit_before("enqueue", message, delay)
+                properties = pika.BasicProperties(
+                    delivery_mode=2,
+                    priority=message.options.get("broker_priority"),
+                )
                 self.channel.basic_publish(
                     exchange="",
                     routing_key=queue_name,
```

2. 

```diff
diff --git a/dramatiq/brokers/rabbitmq.py b/dramatiq/brokers/rabbitmq.py
index af31b83..4314032 100644
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -269,6 +269,12 @@ class RabbitmqBroker(Broker):
             "x-message-ttl": DEAD_MESSAGE_TTL,
         })
 
+    def _build_pika_properties(self, message):
+        return = pika.BasicProperties(
[...]

PR #298, opened 2020-03-30
feat(rabbitmq): add mandatory flag option

davidt99:
Also resolve Bogdanp/dramatiq#290

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -23,6 +23,7 @@ from itertools import chain
 from threading import Event, local
 
 import pika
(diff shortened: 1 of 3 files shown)


Bogdanp:
I haven't been keeping up. What does the `mandatory` flag do, exactly? How is this different from `confirm_delivery`/how does it interact with it?

davidt99:
It makes sure the message reaches a queue oppose to reaches an exchange (which doesn't necessary have a binding to queue).
I do understand that in the current dramatiq implementation it doesn't really make sense but in the near future I want to add exchange support to dramariq so this will be relevant.

PR #303, opened 2020-04-06
Add throws to actors for handling expected exceptions

### LINKED ISSUE #300: Feature Request: Expected errors
I would like to propose adding one more useful feature that we currently use in celery and will help us finish our migration in a straightforward way.

That is of giving the option to pass to the actor a list of exceptions for which:
* No retries will occur
* the exception will be normally stored in the results backend so the caller can get the resulting error.
* Doesn't log the event as error but as info level.

This is actually the equivalent of https://docs.celeryproject.org/en/stable/userguide/tasks.html#Task.throws

This is very fitting for use cases where tasks use parameters or data provided by users, for combinations of whom the task may fail in expected ways (that you can't know itwill happen before running it).

In such cases there is no way this may succeed by retrying, there is no value in polluting your sentry/logs with error level info, and you want the caller side to be able to get the resulting error of this task so it can return the appropriate error response.

This also applies to other types of expected errors.

@Bogdanp whould you integrate this feature? I can work on the implementation and provide a PR

takhs91:
Closes #300

### DIFF
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import inspect
 import re
 import time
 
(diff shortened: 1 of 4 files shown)


Bogdanp:
Thanks! I made some changes on top and then merged this with rebase. I think I will release this and all the other changes next weekend.

Issue #305, opened 2020-04-12
Typing support for the public API


sobolevn:
Hi! Thanks a lot for building this awesome tool!

I am wondering if providing `typing` support is on your list?
It would be great to have `mypy` checked code, especially in async code.

This is one of the main reasons I look for better celery alternatives. They don't have types and there's no plan to support them. It would be awesome to have at least user-facing API typed and [pep-0561](https://www.python.org/dev/peps/pep-0561/) compatible.

I can probably help with adding some basic stuff 👍

Bogdanp:
@davidt99 was asking about the same thing in https://github.com/Bogdanp/dramatiq/issues/302 . I wouldn't mind if someone typed the public API, but I'm not likely to use mypy myself. 

Ideally, the typing would be done in .pyi files to impact the library itself as little as possible. Assuming something like that is possible with modern mypy, then I'd accept such an addition.

davidt99:
I'm not using mypy myself, I just wanted better coding experience. I don't mind adding the type annotations to the codebase itself (not in pyi files) but I'll respect what @Bogdanp decides.

sobolevn:
I am going to send a WIP PR with the type stubs I have created using `monkeytype`.
They are generated using type information from examples. And *does* require further manual editing.

sobolevn:
Well, sadly there are some cases that do not type well (decorators mostly) without a custom `mypy` plugin. How do you feel about that? 😄

Bogdanp:
@sobolevn  Would the plugin have to be distributed with dramatiq? Could a user just install dramatiq and the plugin separately?

sobolevn:
There are two possible way to handle it:
1. Two separate packages: `dramatiq` and `dramatiq-stubs`. The first one is going to be completely untouched. Possibly, adding new optional installation group: `pip install dramatiq[types]`. And the `dramatiq-stubs` repo can contain the full stubs information there is. And custom `mypy` plugins that are also optional. The only problem is that one has to sync two versions of code in two repos. It might be hard.
2. Everything is in one repo. Stubs can be located in the source code (if they are simple enough) or in stubs files in complex cases. `mypy` plugin cam still be optional (sometime soon `mypy` team is going to release new types to modify function signature, but there's not estimate: then we can drop this plugin). In this case it is way easier to sync code and type, but the number of files might get bigger up to 2 times. 

[...]

PR #306, opened 2020-04-12
WIP: adds types for user facing API

sobolevn:
Related #305 

Here's how it works:

```python
from dramatiq import actor

@actor
def test(arg: int) -> str:
    return str(arg)

reveal_type(test)
reveal_type(test(1))
```

Output:

```
» mypy ex.py 
ex.py:7: note: Revealed type is 'dramatiq.actor.Actor[def (arg: builtins.int) -> builtins.str]'
ex.py:8: note: Revealed type is 'builtins.str'
```


sobolevn:
@Bogdanp let's decide what "public API" is. Because project has lots of files and classes / methods. I would be very grateful if you can help me in this field.

sobolevn:
I use `# noqa: B950` because otherwise when I add new lines inside function definition `python3.5` raises parsing errors. I don't know why.

sobolevn:
Self notes: I need to parametrize `Message` as `Generic` on its result. And `get_result()` method will return the generic type. 

This also means, that we need to somehow parametrize the result backend.

Bogdanp:
> @Bogdanp let's decide what "public API" is. 

I would say anything that is documented in the Reference is part of the public API.

dramatiq/__init__.pyi
```diff
@@ -0,0 +1,31 @@
+from .actor import Actor as Actor
```
Bogdanp, line 1: Are the aliases required for some reason?

sobolevn: Yes, `mypy` requires aliases to be configured this way. It is related to `implicit_reexport` setting. And sometimes `mypy` does not tend to see these values with `as` part.

Bogdanp: Got it. Thank you!

sobolevn:
Hi, @Bogdanp! I don't have any more time to work on this.

What is going to be the future of this PR? Looks like it is a good start.

Bogdanp:
I don't really have the time to carry it over either, but I'll happily make any necessary final adjustments and merge it if someone picks it up and finishes it.

dramatiq/__init__.pyi
orsinium: I think mypy should also be fine with it if you provide an explicit `__all__` and include the imported object there.

orsinium:
I'd be happy to take over it. However, since dramatic doesn't support Python 2 anymore, no need to use stubs. Would you mind having type annotations right in the code? That's much easier to maintain than a bunch of duplicated signatures in separate stub files.

Bogdanp:
> I'd be happy to take over it. However, since dramatic doesn't support Python 2 anymore, no need to use stubs. Would you mind having type annotations right in the code? That's much easier to maintain than a bunch of duplicated signatures in separate stub files.
> 
> UPD: now I see, stubs were used only for `__init__.py`. Interesting. I miss the historical context here, but I'm pretty sure it's not needed anymore.

Feel free to pick it up! I have no opinion on the stubs and, if they're no longer necessary, then you should feel free to get rid of them.

orsinium:
Thank you :) Since I work at Sendcloud, I'm sure I can get some time to work on dramatiq. I'll pick it up in one of the upcoming sprints 👍🏿

sobolevn:
@orsinium thanks! 👍

Bogdanp:
I believe this is covered by @orsinium 's recent changes so I'm going to go ahead and close this. Thank you both!

sobolevn:
@orsinium thank you for taking this over! 
Great work!

PR #307, opened 2020-04-12
Adds python3.8 support

sobolevn:
This is an experiment to see if python3.8 is supported

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-18.04"]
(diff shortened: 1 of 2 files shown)


Bogdanp:
3.8 doesn't have pylibmc wheels yet so it's not going to build properly, but dramatiq will run just fine on it. A couple weeks ago I made it install libmemcached-dev from apt and then the build would pass but it would be flaky b/c Azure's apt cache seemed to fail a lot during that time (probably because of the increased network stress given the current situation). Once that gets better, I'll bring those commits back.

sobolevn:
@Bogdanp 👍

PR #310, opened 2020-04-13
Miscellaneous minor changes

dimrozakis:
Hi @Bogdanp 

This PR contains several small changes. Please check the commit log for details. The changes are unrelated but each one is very small and separated into its own commit with descriptive message. Please tell me what you think. If you'd like me to split it into multiple PRs, just say the word.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 /.benchmarks
 /.cache
-/.coverage
+/.coverage*
 /.pytest_cache
 /.tox
 /benchmark*.svg
@@ -9,3 +9,5 @@
 /*.egg-info
 /htmlcov
 __pycache__
+/venv
+.*.sw[pon]
--- a/dramatiq/compat.py
+++ b/dramatiq/compat.py
@@ -39,6 +39,9 @@ class StreamablePipe:
     def fileno(self):
         return self.pipe.fileno()
 
+    def isatty(self):
+        return False
+
     def flush(self):
         pass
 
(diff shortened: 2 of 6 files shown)


Bogdanp:
Thanks! I think I'm only going to keep the `isatty` change. The change to store the exception type and message separately seems redundant with the `exn` field. I'd either want to keep it as-is or just update the `exn` field to make it a dictionary containing `type` and `message` fields.

Bogdanp:
I've merged the isatty change with rebase. Thanks again!

dimrozakis:
Hi @Bogdanp 

Thanks for merging the `isatty` change.

>The change to store the exception type and message separately seems redundant with the exn field. I'd either want to keep it as-is or just update the exn field to make it a dictionary containing type and message fields.

They are redundant. I want to be able to differentiate between different exceptions on the caller's side without having to parse a string to extract exception type name and message. Turning `exn` into a dictionary would be great, the only reason I implemented with additional fields was in order not to break compatibility.

The other changes (.gitignore & compose for tests) aided me in development and I thought of sharing them in case you wanted to include them, they're minor changes and don't really matter anyway.

Bogdanp:
> They are redundant. I want to be able to differentiate between different exceptions on the caller's side without having to parse a string to extract exception type name and message. Turning exn into a dictionary would be great, the only reason I implemented with additional fields was in order not to break compatibility.

Yup. I figured. I don't think we have to worry about breaking compatibility since the change hasn't been officially released yet so we can definitely make the change now. Let me know if you want to do it yourself.

PR #315, opened 2020-06-01
Avoid unnecessary calls to declare_queue on RabbitmqBroker.enqueue()

AndreCimander:
## Summary

Enqueuing tasks results in unnecessary calls to declare_queue, this patch aims to avoid those calls (and also adds a minor function call save in Actor.message_with_options).

It's a minor improvement performance-wise, but saving energy is always nice and good for the environment :) 

### Adding 10,000 tasks:

`2644297 function calls (2644200 primitive calls) in 1.940 seconds` (with patch)

vs

`2704295 function calls (2704198 primitive calls) in 2.061 seconds` (without patch)

### DIFF
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -85,6 +85,8 @@ class Actor:
         """
         for name in ["on_failure", "on_success"]:
             callback = options.get(name)
+            if callback is None:
+                continue
             if isinstance(callback, Actor):
                 options[name] = callback.actor_name
 
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -282,7 +282,8 @@ class RabbitmqBroker(Broker):
             has been closed.
         """
         queue_name = message.queue_name
-        self.declare_queue(queue_name, ensure=True)
+        if queue_name not in self.queues or queue_name in self.queues_pending:
+            self.declare_queue(queue_name, ensure=True)
 
         properties = pika.BasicProperties(
             delivery_mode=2,


dramatiq/actor.py
```diff
@@ -85,6 +85,8 @@ def message_with_options(self, *, args=None, kwargs=None, **options):
         """
         for name in ["on_failure", "on_success"]:
             callback = options.get(name)
+            if callback is None:
```
Bogdanp, line 88: Did you intend to keep this change here?

AndreCimander: It's a minor optimization that skips the call overhead of the `if...elif`. IIRC it was ~30% of the saved function calls. Granted, it will will add overhead for cases were both callbacks are supplied, which I assumed is rarely the case.

Your call if you want to keep it, I'll update the PR accordingly :)

Bogdanp: Let's drop it. I don't imagine it made up 30% of the total time so it's probably not worth optimizing compared to `declare_queue` which involves IO.

AndreCimander: I pulled the profile code from my remote machine (note to self, sync scratch files):

2704297 function calls (without patch)
2684297 function calls (with patch but without continue)
2644297 function calls (with patch and continue)

So it's actually 2/3 of the saved calls, oof. Let me check the overhead for having one or both callbacks defined, brb.

AndreCimander: # one callback defined (method)
2644297 calls (with patch)
2684297 calls (with patch but without continue)

# both callbacks defined (method)
2644297 calls (with patch)
2684297 calls (with patch but without continue)

Interesting, looks like the compiler is already doing some optimizations, so no overhead for the additional continue conditional when the callbacks are not None.

Bogdanp: Yes, I can imagine there would be many calls here, but calls by themselves don't concern me that much. What are the timings like? My guess is the vast majority of the time is being spent in the declare_queue calls rather than here.

AndreCimander: It's 6ms/0.4% (1.1% instead of 1.5%) for 10k tasks. I've attached the call graphs for both version ((offtopic to this PR, switching to [orjson](https://github.com/ijl/orjson) might offer some nice gains)) :

Screenshot_20200606_202458_without_continue
Screenshot_20200606_202545_with_continue

PR #316, opened 2020-06-01
Facilitate overwriting of Consumer class

AndreCimander:
This is just a very minor nitpick, feel free to close it, but it would be nice to change the class of the Consumer without the need to completly replicate and overwrite `.consume()` :)

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -109,6 +109,7 @@ class RabbitmqBroker(Broker):
         self.queues = set()
         self.queues_pending = set()
         self.state = local()
+        self.consumer = _RabbitmqConsumer
 
     @property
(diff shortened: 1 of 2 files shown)


Bogdanp:
This seems OK to me, but how about making it a class property instead?

AndreCimander:
I was thinking about adding it as a cls property, but the consumer classes are not yet defined.


We could go the `@property` route, though, this would also make it easier to add logic if one desires. How would you like it named? `consumer` / `consumer_class`? :)


```
    @property
    def consumer_class(self):
        return _RedisConsumer
```

Bogdanp:
Ah, right, that makes sense. Yeah, `consumer_class` sounds good.

Bogdanp:
Thanks! I have squashed and merged this.

PR #317, opened 2020-06-07
Proposal: build optimized middleware lists per signal, don't execute empty default methods

AndreCimander:
After digging a little deeper into the call graph I identified some additional calling overhead that can be optimized :slightly_smiling_face: 

#### Benchmark (adding 10,000 tasks):

without patch:
2704295 function calls(2704198 primitive calls) in 1.925 seconds

with patch:
2474322 function calls (2474225 primitive calls) in 1.713 seconds

### DIFF
--- a/dramatiq/broker.py
+++ b/dramatiq/broker.py
@@ -85,12 +85,11 @@ class Broker:
             self.add_middleware(m)
 
     def _build_optimized_middleware_list(self, signal: str):
-        self.middleware_by_signal[signal] = optimized_list = list()
-        for middleware in self.middleware:
-            middleware_handler = getattr(middleware, signal, False)
-            if middleware_handler is False or hasattr(middleware_handler, "empty"):
-                continue
-            optimized_list.append(middleware_handler)
+        self.middleware_by_signal[signal] = [
+            getattr(middleware, signal)
+            for middleware in self.middleware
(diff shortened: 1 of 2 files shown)


Bogdanp:
Thanks! I'll echo my [other comment](https://github.com/Bogdanp/dramatiq/pull/315#issuecomment-640471128) here to say that I think a better approach would be to come up with a bulk enqueue API rather than try to optimize things at this level.

AndreCimander:
@Bogdanp no problem :+1:

But I would like to bring forward the argument that this will not just improve enqueuing times, but almost all signal handling across the board (ack, nack, declare_actor, declare_queue, delay_message, process_message), improving startup times and message throughput.

As a side note, improving energy efficiency and reducing the CO2 footprint should be an important focus for every developer, especially for software that might power lots of infrastructure (and I see dramatiq as the go-to replacement for Celery).

Using `__qualname__` would also eliminate the need to touch the base middleware:

```
    def _build_optimized_middleware_callable(self, signal: str):
        self.middleware_by_signal[signal] = [
            getattr(middleware, signal)
            for middleware in self.middleware
            # don't add no-op base middleware methods
            if getattr(middleware, signal).__qualname__[0:11] != 'Middleware.'
        ]
```

But no worries, I'll use this in my projects anyway  :slightly_smiling_face:

PR #318, opened 2020-06-08
ci: add 3.8 to build matrix

Bogdanp:
Blocked by a lack of memcached wheels for 3.8.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@master
       - uses: actions/setup-python@v1
         with:
-          python-version: "3.7"
+          python-version: "3.8"
       - run: pip install tox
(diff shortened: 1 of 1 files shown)


davidt99:
@Bogdanp  Can we skip memcache for 3.8? Python 3.9 is expected to be released in a couple of months...

Bogdanp:
We'd have the same problem with CI on 3.9. Looks like the [pylibmc](https://github.com/lericson/pylibmc/pull/257) maintainer hasn't published changes in a while. The simplest route might be to just install the prerequisites to build libmemcached in CI. Alternatively, someone could publish a pylibmc fork to pypi with wheels for 3.8 and 3.9. I'll take another look at this when I can. Happy to take help if someone wants to get the build passing (you probably just need to install the right development library (libmemcached-dev) and a couple dependencies).

Issue #320, opened 2020-06-17
redis.exceptions.ResponseError: @user_script:218: @user_script: 218: Lua redis() command arguments must be strings or integers


vkatochka:
Dramatiq 1.9.0
Redis server v=5.0.7, redis-cli 5.0.7
Ubuntu 18.04.4 LTS
Python 3.6.9

~~~
@dramatiq.actor(max_retries=0,max_age=((2*60*1000)+10000),time_limit=((2*60*1000)+10000))
def mkss(sid,uuid,prms):
	...
 ~~~

~~~
[2020-06-17 07:30:48,475] [PID 2044] [Thread-4] [dramatiq.middleware.age_limit.AgeLimit] [WARNING] Message 'unique-1592378746761-5ee9c57a55536e1fe1518ad6' has exceeded its age limit.
[2020-06-17 07:30:48,477] [PID 2044] [Thread-4] [dramatiq.worker.ConsumerThread(default)] [ERROR] Unhandled error during post_process_message(mkss('5ee9c57a55536e1fe1518ad6', 'unique-1592378746761-5ee9c57a55536e1fe1518ad6', {...})).  You've found a bug in Dramatiq.  Please report it!
Although your message has been processed, it will be processed again once this worker is restarted.
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/dramatiq/worker.py", line 335, in post_process_message
    self.consumer.nack(message)
  File "/usr/local/lib/python3.6/dist-packages/dramatiq/brokers/redis.py", line 317, in nack
    self.broker.do_nack(self.queue_name, message.options["redis_message_id"])
  File "/usr/local/lib/python3.6/dist-packages/dramatiq/brokers/redis.py", line 275, in do_dispatch
    return dispatch(args=args, keys=keys)
  File "/usr/local/lib/python3.6/dist-packages/redis/client.py", line 4073, in __call__
    return client.evalsha(self.sha, len(keys), *args)
  File "/usr/local/lib/python3.6/dist-packages/redis/client.py", line 3143, in evalsha
    return self.execute_command('EVALSHA', sha, numkeys, *keys_and_args)
  File "/usr/local/lib/python3.6/dist-packages/redis/client.py", line 901, in execute_command
    return self.parse_response(conn, command_name, **options)
  File "/usr/local/lib/python3.6/dist-packages/redis/client.py", line 915, in parse_response
    response = connection.read_response()
[...]

PR #323, opened 2020-06-27
Replace `message ~= nil` in dispatch.lua, since a negative result returns false and not nil

swidoff:
This is a fix for [issue 320](https://github.com/Bogdanp/dramatiq/issues/320).

Apparently, `redis.call("hget"...)` doesn't return nil if the key does not exist, it returns false.

Here's an example script I wrote:

```
local queue = "test-queue"
local message_id1 = "key"
local message_id2 = "not-a-key"

redis.call("hset", queue, message_id1, "value")

local message1 = redis.call("hget", queue, message_id1)
local message2 = redis.call("hget", queue, message_id2)

redis.debug(message1)
redis.debug(message2)

if message1 ~= nil then
   redis.debug("Message 1 is not nil with ~=")
end

if message1 then
   redis.debug("Message 1 is not nil")
end

if message2 ~= nil then
   redis.debug("Message 2 is not nil with ~=")
end

if message2 then
   redis.debug("Message 2 is not nil")
end
```

The output is:

```
 redis-cli --ldb --eval test.lua 
Lua debugging session started, please use:
quit    -- End the session.
restart -- Restart the script in debug mode again.
help    -- Show Lua script debugging commands.

14882:C 27 Jun 2020 00:38:42.506 # Redis forked for debugging eval
* Stopped at 1, stop reason = step over
-> 1   local queue = "test-queue"
lua debugger> c
14882:C 27 Jun 2020 00:38:42.977 # Lua debugging session child exiting
<debug> line 10: "value"
<debug> line 11: false
<debug> line 14: "Message 1 is not nil with ~="
<debug> line 18: "Message 1 is not nil"
<debug> line 22: "Message 2 is not nil with ~="
```

The second `hget` of an unmapped key returns false and so the result will always be not nil.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -42,3 +42,4 @@ of those changes to CLEARTYPE SRL.
 | [@benekastah](https://github.com/benekastah)          | Paul Harper            |
 | [@timdrijvers](https://github.com/timdrijvers)        | Tim Drijvers           |
 | [@takhs91](https://github.com/takhs91)                | Takis Panagopoulos     |
+| [@swidoff](https://github.com/swidoff)                | Seth Widoff            |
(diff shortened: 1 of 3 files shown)


Bogdanp:
Thanks! Great work tracking this down. I've now merged your changes with rebase.

PR #325, opened 2020-07-10
Minor change to the docs on enqueuing Redis messages from other languages.

swidoff:
Clarify that a `redis_message_id` must be set in the `options` field of the message and that that ID is the one used in the `HSET` and `RPUSH`.


Bogdanp:
Thanks!

PR #328, opened 2020-07-13
Raise an error if watch is used on an unsupported system.

### LINKED ISSUE #326: Document that watch is not supported on windows
# Issues

Watch requires the signal `SIGHUP` that is only available on unix type system.
## What OS are you using?
windows
1.9.0
## What did you do?
started dramatiq with the --watch option and modified the files causing a reload
The code would reload. Or a meaningful error would be raised.
## What happened?
The excection `AttributeError: module 'signal' has no attribute 'SIGHUP'` is raised.

While it would be nice to have support for reload on windows, I think its would be ok to just document that watch is not supported in `dramatiq --help` and maybe ignore the flag on windows or raise a more meaningful exception than the attribute error

CaselIT:
Fixes #326

I'm not sure if / how should this be tested. Currently all cli test are skipped on windows.


CaselIT:
🥂

PR #330, opened 2020-07-15
Worker timeout option

mic47:
Currently when doing graceful shutdown of dramatiq, there is 10 minutes timeout for workers. That means that any worker computing longer than 10 minutes will be terminated. This PR adds an option to set this timeout on command-line, so longer workloads will not be terminated by graceful shutdown.


mic47:
For some reason, CI test fails on 600_000 even though I fixed it...

Bogdanp:
Thanks! I've renamed the flag to `worker-shutdown-timeout` for consistency and merged the changes with rebase.

PR #332, opened 2020-08-08
Don't allow Result middleware without a result backend.

AndreCimander:
Since I saw some issues related to the result middleware and django_dramatiq lately, I think it might be sensible to be more strict and descriptive when using the result middleware the wrong way.

My first idea was to use a fallback call like `get_broker`, but this would also require changes to `django_dramatiq` which needs to set the backend first, but it's currently **way** too hot to be really productive and motivated, so raising an descriptive error was my next idea :upside_down_face: 

Feel free to adjust the error message, on a second read it sounds a bit harsh.

### DIFF
--- a/dramatiq/results/backend.py
+++ b/dramatiq/results/backend.py
@@ -21,7 +21,7 @@ import typing
 
 from ..common import compute_backoff, q_name
 from ..encoder import Encoder
-from .errors import ResultMissing, ResultTimeout, ResultError
+from .errors import ResultMissing, ResultTimeout
 from .result import unwrap_result, wrap_exception, wrap_result
 
 #: The default timeout for blocking get operations in milliseconds.
--- a/dramatiq/results/backend.py
+++ b/dramatiq/results/backend.py
@@ -21,7 +21,7 @@ import typing
 
 from ..common import compute_backoff, q_name
 from ..encoder import Encoder
-from .errors import ResultMissing, ResultTimeout
+from .errors import ResultMissing, ResultTimeout, ResultError
 from .result import unwrap_result, wrap_exception, wrap_result
 
 #: The default timeout for blocking get operations in milliseconds.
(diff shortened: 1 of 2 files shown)


Bogdanp:
I'm reluctant to make this change since it's technically breaking and, if we were to go in this direction, then I'd rather just make the argument required than have it be optional and error out when it's `None`. I'll have to think about this some more.

AndreCimander:
Yeah, I have also thought about making the backend parameter required, since the middleware is technically broken without a backend, but you'll lose the ability to give a more specific error message, eventually leading to more confusion for `django_dramatiq`  users.

Maybe just changing `django_dramatiq`'s behaviour to automatically inject the backend into any result-middlewares without a backend is the better approach, but this still has the potential to lead to a broken middleware state if one adds the middleware manually without a backend :thinking:

jenstroeger:
I’ve been thinking about a similar/related scenario: in a codebase I worked with there are oodles of actor functions, some of which return values and some don’t. Without a result backend Dramatiq’s worker issues a warning:

https://github.com/Bogdanp/dramatiq/blob/8dd6a752c63cf89d85a1b265ca707c233b810814/dramatiq/worker.py#L478-L486

Which made me wonder: would you be open to a “NullBackend” which simply ignores results altogether? A bit like the [StubBackend](https://dramatiq.io/reference.html#dramatiq.results.backends.StubBackend). Then we’d (temporarily) add the middleware and drop the result values, maybe log the actor function that attempted to return a result. The NullBackend would be a nice addition to the Dramatiq [result backends](https://github.com/Bogdanp/dramatiq/tree/master/dramatiq/results/backends)?

jenstroeger:
> @jenstroeger NullBackend functionality can be achieved through extending the existing backend class, but as stated above we won't be able to add a new backend to the core. I don't think there were any requests for a such implementation.

No worries. I’ve been running such a NullBackend fine and as part of the general setup in our context, even though it’s not supplied by the Dramatiq package 👍🏼

Issue #333, opened 2020-08-19
cli returns RET_OK when subprocess killed with SIGKILL


omegacoleman:
# Issues

When subprocess killed with SIGKILL, the main process exit with code 0, falsely indicating normal exit.

## Checklist

* [x] Does your title concisely summarize the problem?
* [x] Did you include a minimal, reproducible example?
* [x] What OS are you using?
* [x] What version of Dramatiq are you using?
* [x] What did you do?
* [x] What did you expect would happen?
* [x] What happened?

## What OS are you using?

MacOSX 10

## What version of Dramatiq are you using?

trunk version from github master

## What did you do?

first run dramatiq cli with subprocess (any broker is okay):
```
dramatiq --processes 8 ...
```

get a subprocess PID via verbose log or htop,and kill it with sigkill:
```
kill -9 XXXXX(the subprocess PID)
```

the main process was able to detect that and shut down:
```
[CRITICAL] Worker with PID 88666 exited unexpectedly (code -9). Shutting down...
```

however the exit code is RET_OK==0:
```
$ echo $?
0
```

## What did you expect would happen?

The exit code should be RET_KILLED==1 in this case:
```
$ echo $?
1
```

## What happened?

The subprocess kill status is eaten by main process, and returns 0.

This caused issue in our environment (we used dramatiq with supervisor), the supervisor was baffled that it thinks the process exited normally and didn't try to restart (it's OOM killed indeed).

PR #334, opened 2020-08-19
return RET_KILLED when subprocess is terminated by unhandlable signal

omegacoleman:
return RET_KILLED when subprocess is terminated by unhandlable signal (like SIGKILL)

fixes https://github.com/Bogdanp/dramatiq/issues/333


Bogdanp:
Looks great, thanks!

PR #341, opened 2020-09-30
ci: Add Python 3.8

rouge8:
Closes #318.

I also did some tox cleanup so that we wouldn't need to install libmemcached-dev to run `tox -e lint`.

The Windows changes are because of https://docs.python.org/3.8/whatsnew/3.8.html#bpo-36085-whatsnew


Bogdanp:
This is great, thank you!

Issue #342, opened 2020-10-01
Actor in pipeline will not ignore output of previous actor.


jtfidje:
## What OS are you using?

Debian 10

## What version of Dramatiq are you using?

1.9.0

## What did you do?

I followed the documentation to have one actor in a pipeline ignore output of previous actor. 
Below are two things I tried:

```
pipeline([
    send_extraction_request.message(task_id, _url, payload),
    get_extraction_status.message_with_options(task_id, _url, pipe_ignore=True)
]).run()
```

```
pipeline([
    send_extraction_request.message(task_id, _url, payload),
    get_extraction_status.message_with_options((task_id, _url), pipe_ignore=True)
]).run()
```

## What did you expect would happen?

I expected `get_extraction_status` to get run with `task_id`, and `_url`, but ignore the output from `send_extraction_request`

## What happened?

I get the following error message for the two things I tried:

`TypeError: message_with_options() takes 1 positional argument but 3 were given`  

`TypeError: message_with_options() takes 1 positional argument but 2 were given`

## Solution
I found out how to do this, and have raised a PR to change the documentation that reflects how this should be done with the current codebase. 
If my solution is wrong, please let me know how I should actually be doing this :-)

Bogdanp:
The right syntax here should be:

```
pipeline([
    send_extraction_request.message(task_id, _url, payload),
    get_extraction_status.message_with_options(args=(task_id, _url), pipe_ignore=True)
]).run()
```

If that doesn't work for some reason, feel free to reopen the ticker.

PR #343, opened 2020-10-01
docs: Update docs to reflect current parameter requirements

jtfidje:
This PR changes the documentation to fix Issue #342


Bogdanp:
Thanks, but I think this may end up confusing people more. The right way to pass an argument is to pass `args=(1,)`, but both `args` and `kwargs` are optional.

jtfidje:
Thanks for looking into this so quick :)

PR #354, opened 2020-10-16
Update rabbitmq.py

evstratbg:
support `dramatiq_dead_message_ttl` env variable to dynamically change it


Bogdanp:
Thanks. It looks like there are linting errors here and I think the types don't line up. Doesn't the env var need to be converted to an integer?

evstratbg:
> Thanks. It looks like there are linting errors here and I think the types don't line up. Doesn't the env var need to be converted to an integer?

Right. Should I add more complex checks or just add int()?

Bogdanp:
Just int is fine, thanks!

Bogdanp:
Thanks, this has been merged with rebase.

PR #356, opened 2020-10-18
Drop support to Python 3.5

jairhenrique:


Bogdanp:
Thanks, but it's a little too early for this. I'll do this when I start working on the 2.0  branch. Having Python 3.5 support in the mean time isn't impacting things much.

Issue #370, opened 2020-12-18
Windows CI broken


Bogdanp:
https://github.com/Bogdanp/dramatiq/runs/1559303340 looks like CI on Windows has been broken for a while. memcached is either not starting properly or failing to bind to the right port.

It'd be a great help if someone with Windows experience could give this a look. Otherwise, I'll probably get to it eventually.

PR #371, opened 2020-12-20
ci: fix memcached for windows

gdvalle:
This addresses broken tests in the CI env by using a native port of
memcached for the test suite... I'm not sure if it's crashing, or unable
to handle the load from the test suite, but the native copy seems to
do fine.

Fixes https://github.com/Bogdanp/dramatiq/issues/370.


Bogdanp:
This is great. Thanks, Greg!

PR #372, opened 2020-12-21
logging: omit result warning with piped messages

gdvalle:
This avoids the Results middleware's warning about a return value that
isn't stored when piped messages are used.

The actor's warning about returning results already has this behavior,
but it felt like we should avoid the log message for piped messages w/
Results middleware too.


gdvalle:
Woops.  Force-pushed for a linting issue there 😄

Bogdanp:
Thanks!

Issue #213, opened 2019-05-16
Retries middleware options are ignored if task receives SIGKILL or equivalent while running


MihaiBalint:
# Issues

Steps to reproduce:
1. Create a task that does something insignificant (for .e.g sleeps for 5 mins)
2. Configure the task with max_retry=3 times. Start the worker and the task.
3. Find the task pid and send SIGKILL to it. Notice that the task is restarted immediately
4. Repeat this 5 or more times and notice that the task is restarted immediately every time even after the the number of retries exceeds the configured max_retry.

In production the process could be killed because the task ran out of memory and the kernel OOM selected to kill it.

This could also be automated by having a task that simply performs suicide using something like this: os.kill(os.getpid(), signal.SIGKILL).

I would expect the max_retry and other options from the Retires middleware to also work if the task process is terminated abruptly.

## What OS are you using?

Using Linux, Ubuntu 18.04.

## What version of Dramatiq are you using?

dramatiq 1.6.0

Bogdanp:
This is to be expected and the behavior you're seeing has nothing to do with the Retries middleware.

Dramatiq has at-least-once message processing semantics.  When a message is received from the broker, it is put in a special lease state on the broker and only after the message is successfully processed by a worker is it removed from the broker.  If the Dramatiq worker that is processing said message is killed before it can remove the message (i.e. "acknowledge" that is has been processed), then the broker will expire the lease and put the message back into circulation for another worker to process it at a later time.

If we were to implement some sort of eager-acknowledgement functionality (we won't, because that's basically never what you want), the result would be even worse.  The message would be received from the broker and immediately acknowledged, the worker would then be killed *while* processing the message. Since the worker has been killed, there's no one to run the `Retries` middleware and the message has been deleted from the broker (due to us acknowledging it early).  Poof goes the message!

As far as I'm aware, there is no way to tell RMQ that a message can only be leased a certain number of times and, while adding this to the Redis broker would be possible, I think it would introduce a lot of unnecessary complexity.  The solution to your problem (assuming it's the OOM-killing you mentioned) is to track down whatever is causing the memory leaks or to provision larger machines to run your workers.

george-miller:
@Bogdanp 
[...]

PR #383, opened 2021-02-10
Record when a message is recovered on a killed worker, and don't retry those tasks infinitely

george-miller:
This is a POC for the feature of handling worker OOMKills.  I have started discussions  https://github.com/Bogdanp/dramatiq/issues/213 and https://www.reddit.com/r/dramatiq/comments/lgyhvl/why_is_do_maintenance_not_locked/? related to this, but I think this is a great solution, it works for me locally very well.  Obviously this PR isn't done, but please give me your thoughts!  Is this something we can consider adding once it's more fully written out?  Should I make it a separate middleware?

### DIFF
--- a/dramatiq/brokers/redis/dispatch.lua
+++ b/dramatiq/brokers/redis/dispatch.lua
@@ -116,18 +116,13 @@ if do_maintenance == "1" then
         local message_ids = redis.call("smembers", dead_worker_queue_acks)
         if next(message_ids) then
             for i, message_id in ipairs(message_ids) do
-              local encoded_message = redis.call("hget", queue_messages, message_id)
-              if encoded_message ~= nil then
-                local message = cjson.decode(encoded_message)
-                -- Do we need to do this in case of multiple people running domaintenance?
-                -- redis.call("hdel", queue_messages, message_id)
-                if message['options']['revives'] ~= nil then
-                  message['options']['revives'] = message['options']['revives'] + 1
-                else
-                  message['options']['revives'] = 0
-                end
-                redis.call("hset", queue_messages, message_id, cjson.encode(message))
+              local message = cjson.decode(redis.call("hget", queue_messages, message_id))
+              if message['options']['revives'] ~= nil then
+                message['options']['revives'] = message['options']['revives'] + 1
+              else
+                message['options']['revives'] = 1
               end
+              redis.call("hset", queue_messages, message_id, cjson.encode(message))
             end
 
             for message_ids_batch in iter_chunks(message_ids) do
--- a/dramatiq/brokers/redis/dispatch.lua
+++ b/dramatiq/brokers/redis/dispatch.lua
@@ -115,6 +115,21 @@ if do_maintenance == "1" then
         local dead_worker_queue_acks = dead_worker_acks .. "." .. queue_name
         local message_ids = redis.call("smembers", dead_worker_queue_acks)
         if next(message_ids) then
+            for i, message_id in ipairs(message_ids) do
+              local encoded_message = redis.call("hget", queue_messages, message_id)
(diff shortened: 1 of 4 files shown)


Bogdanp:
Thanks for the PR, but I don't think this is the right approach for a few reasons:

* it's Redis-only and, AFAIK, could not be changed to support RMQ (which doesn't allow you to update enqueued messages),
* it introduces (likely) unacceptable overhead by parsing and encoding the message JSON inside Redis and
* it can cause unexpected data loss.

My view is still that the right fix in cases like these is to improve your monitoring (i.e. track memory usage across workers) and fix any memory leaks you find. Leaks are bugs and they shouldn't be swept under the rug like this.

george-miller:
@Bogdanp 

I agree leaks are bugs.  But not all oomkills or worker kills are leaks.  Some oomkills can be an unexpected valid memory increase, or simply a misconfiguration of resource limits.  In my case, we use dramatiq to run queries on a variety of datasources (qubole, databricks, mysql, graphite, etc).  When we go to fetch the results, we don't always know exactly how big the results set is going to be.  Thus at the end of the task run there's a massive memory increase which we fetch the results and put them in our cache.  We're going to add in a check before we fetch results and that should solve most problems but some datasources don't have that api.  If you know of any other task runner or pattern that would be better for us to use to run queries asynchronously, please lmk.

Either way, to infinitely retry a task that causes the worker to die is not a good strategy, no matter why the worker died (oomkill, moving nodes, whatever).  Also, wouldn't it be nice to export metrics about workers that died and what tasks were recovered from those workers?  I'm surprised you don't think there's any use in making the do_maintenance process more feature full.

`It introduces (likely) unacceptable overhead by parsing and encoding the message JSON inside Redis`  What performance standards are you striving towards?  We can easily put it behind a flag that's passed into dispatch if that's a concern.  If track_revives is false, don't parse the json at all.  Also if do_maintenance is behind a lock, it shouldn't matter it if takes 10ms longer.

`it can cause unexpected data loss.` Why?  I must be missing something with how this whole thing works.

Bogdanp:
>  When we go to fetch the results, we don't always know exactly how big the results set is going to be. Thus at the end of the task run there's a massive memory increase which we fetch the results and put them in our cache. We're going to add in a check before we fetch results and that should solve most problems but some datasources don't have that api. If you know of any other task runner or pattern that would be better for us to use to run queries asynchronously, please lmk.

The standard solution for these sorts of problems is to read the data in chunks and discard anything that gets too "too big" (whatever that means in the context of the hw you're running). 

> I'm surprised you don't think there's any use in making the do_maintenance process more feature full.

More features means more things that can go wrong and more things to maintain. There are a number of mechanisms to figure out when a worker dies already (hooks, exit codes) and I don't think we need another.

> it shouldn't matter it if takes 10ms longer

A quick test on my machine shows the Redis backend can currently process about 13k messages per second. Adding 10ms of overhead per maintenance would destroy throughput. What's worse, though, is the overhead can vary based on the size of the message. We should be working towards reducing overhead, not adding it.

> Why? I must be missing something with how this whole thing works.

Automatically dropping messages is at odds with Dramatiq's at-least-once semantics. Also, folks can use custom encoders; the code as written would crash if it encountered a pickled message.

If you absolutely want to go this route, I'd recommend forking the Redis broker into your app, but this sort of approach is just not right for Dramatiq.

george-miller:
Makes sense, I see where you're coming from.  

I didn't realize there were hooks for worker_shutdown, that might actually be exactly what I need.  I could separately store revives and `srem` from the worker acks if they get past a certain point so maintenance doesn't revive them ... still kinda a hack but it would allow us to get rid of the infinite retry problem.  I'll give that a try.

Issue #266, opened 2020-01-15
[CRITICAL] Consumer encountered an unexpected error, AttributeError: 'NoneType' object has no attribute 'decode'


vkatochka:
dramatiq 1.7.0

> lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.3 LTS
Release:        18.04
Codename:       bionic

> dramatiq bwfut -Q home -p1 -t1 --pid-file /var/www/.../tmp/dsrv-home-bwfut.pid --log-file /var/www/.../logs/dsrv-home-bwfut.log --path /var/www/.../ & disown

~~~
[2020-01-15 22:39:17,008] [PID 17861] [MainThread] [dramatiq.MainProcess] [INFO] Sending signal 1 to worker processes...
[2020-01-15 22:39:17,011] [PID 17867] [MainThread] [dramatiq.WorkerProcess(0)] [INFO] Stopping worker process...
[2020-01-15 22:39:17,012] [PID 17867] [MainThread] [dramatiq.worker.Worker] [INFO] Shutting down...
[2020-01-15 22:39:18,127] [PID 17867] [MainThread] [dramatiq.worker.Worker] [INFO] Worker has been shut down.
[2020-01-15 22:39:18,260] [PID 17861] [MainThread] [dramatiq.MainProcess] [INFO] Dramatiq '1.7.0' is booting up.
[2020-01-15 22:39:18,280] [PID 19075] [MainThread] [dramatiq.WorkerProcess(0)] [INFO] Worker process is ready for action.
[2020-01-15 22:43:48,642] [PID 17861] [MainThread] [dramatiq.MainProcess] [INFO] Sending signal 1 to worker processes...
[2020-01-15 22:43:48,643] [PID 19075] [MainThread] [dramatiq.WorkerProcess(0)] [INFO] Stopping worker process...
[2020-01-15 22:43:49,357] [PID 19075] [MainThread] [dramatiq.worker.Worker] [INFO] Shutting down...
[2020-01-15 22:43:50,450] [PID 19075] [MainThread] [dramatiq.worker.Worker] [INFO] Worker has been shut down.
[2020-01-15 22:43:50,613] [PID 17861] [MainThread] [dramatiq.MainProcess] [INFO] Dramatiq '1.7.0' is booting up.
[...]

PR #394, opened 2021-04-17
fix(consumer): Fix exception `AttributeError: 'NoneType' object has no attribute 'decode'`

swaroopch:
Seeing this issue in production, which leads to `[CRITICAL] Consumer encountered an unexpected error.` in the logs, which leads to unconsumed messages in the queue ☹️  :

Screen Shot 2021-04-17 at 12 29 30 PM

Possibly related: https://github.com/Bogdanp/dramatiq/issues/266

OS and Python version: [python:3.6-buster](https://hub.docker.com/_/python)

Dramatiq version: `dramatiq[redis]==1.10.0`


Bogdanp:
Thanks! But I think #395 has a higher chance of being the right fix here.

Issue #404, opened 2021-06-07
How to access redelivered Flag available in `RabbitMQConsumer` object


nffdiogosilva:
# Issues

I'm trying to handle some redeliver pitfalls with rabbitMQ, namely having a consumer that exists before sending an acknowledgment. There's a redelivered attribute available in the `_RabbitmqConsumer`, but it doesn't seem accessible in a higher layer of the dramatiq API, meaning my solution would have to be to override the `_RabbitmqConsumer` as well as the `_RabbitmqMessage` to pass the `redelivered` value.

Is there any easier way to know if a message is being re-delivered?

## Checklist

* [ ] What version of Dramatiq are you using?

`1.8.1`

PR #405, opened 2021-06-07
add redelivered attribute to message object

nffdiogosilva:
I think something like this would fix issue #404


Bogdanp:
Yes, I think this would be fine to merge. I'll probably do it this weekend.

Bogdanp:
Thanks! This is now merged w/ rebase.

Issue #389, opened 2021-03-18
Feature Request: Standard way to inject arguments


AstraLuma:
(sorry for ditching the template)

Things I'm wanting to do with dramatiq involve things shaped like dependency injection: the actor requests a kind of dependency (like a progress status callable or a database connection), and then a middleware accepts the flag and injects the dependency into the message.

This is currently not as easy as it could be, because the message is then re-serialized and dependencies are often not serializable (nor should they be included).

Current solution:

```python
class Injector(dramatiq.Middleware):
    @property
    def actor_options(self):
        return {"give_me_the_thing"}

    def before_process_message(self, broker, message):
        actor = broker.get_actor(message.actor_name)
        if actor.options.get("give_me_the_thing", False):
            message.kwargs['the_thing'] = ...

    def after_process_message(self, broker, message, *, result=None, exception=None):
        try:
            del message.kwargs['the_thing']
        except KeyError:
            pass

    after_skip_message = after_process_message
```

hmoravec:
I used this approach for injecting unit of work instance but failed messages were not retried then. Dramatiq emitted standard "Failed to process" message and silently removed it from the queue. However it worked when injecting e.g. integer value. (I added this middleware before Retries middleware.)

AstraLuma:
Yeah, sounds like Dramatiq couldn't re-serialize the `Message`.

That's why I delete the additions afterwards.

Bogdanp:
I think the right level at which to do this is a custom `actor` decorator that wraps dramatiq's and provides DI. Here's an implementation I had for APIStar:

https://github.com/Bogdanp/apistar_dramatiq/tree/42627a5587639bc3dd7c899328782cbfd95b551b/apistar_dramatiq

I don't think I'd want to standardize something like this in Dramatiq, though, because there are many ways you could want to do DI and Dramatiq already provides enough configuration knobs for you do be able to do it yourself.

x0day:
> Yeah, sounds like Dramatiq couldn't re-serialize the `Message`.
> 
> That's why I delete the additions afterwards.

maybe we can re-serialize message like this https://github.com/Bogdanp/dramatiq/issues/366, use `Message.copy` method.

re-serialize message is really useful

PR #406, opened 2021-06-10
Add injections field to Actor and Message

AstraLuma:
This implements #389: provides a way for middleware to add keyword arguments that won't get serialized.

The way I set this up, caller arguments override message injections, which override actor injections.

This is deliberately meant to be a very minimal addition.


AstraLuma:
@Bogdanp if these additions are acceptable, I'll happily write the documentation to go with.

Bogdanp:
Yes, but I don't have time right now and probably won't for a while. Sorry!

Issue #403, opened 2021-06-01
dramatiq-gevent TimeLimitExceeded exception fails to set


FinnLidbetter:
Raising thread exceptions fails when running workers with dramatiq-gevent. This affects the TimeLimitExceeded middleware (and anything using dramatiq.middleware.threading.raise_thread_exception). CRITICAL log lines are logged when the time limit is exceeded and the exception fails to set. Is this expected behaviour for running workers with gevent?

## Checklist

* [x] Does your title concisely summarize the problem?
* [x] Did you include a minimal, reproducible example?
* [x] What OS are you using?
* [x] What version of Dramatiq are you using?
* [x] What did you do?
* [x] What did you expect would happen?
* [x] What happened?

## What OS are you using?
Ubuntu 18.04

## What version of Dramatiq are you using?
1.11.0

## What did you do?
Run workers with dramatiq-gevent and TimeLimitExceeded middleware on a task that may take longer than the time limit. 

## What did you expect would happen?
If the task takes longer than the time limit, the task should timeout and the TimeLimitExceeded exception should be raised in the worker thread the next time that the thread acquires the GIL.

## What happened?
The exception fails to set.
```
[2021-06-01 11:16:26,234] [PID 11500] [Thread-1] [dramatiq.middleware.time_limit.TimeLimit] [WARNING] Time limit exceeded. Raising exception in worker thread 140492336214088.
[2021-06-01 11:16:26,235] [PID 11500] [Thread-1] [dramatiq.middleware.threading] [CRITICAL] Failed to set exception (TimeLimitExceeded) in thread 140492336214088.
```

## What gevent version are you using?
I am using the latest version, 21.1.2, but the issue also happens with older versions of gevent. 

## Reproducible example
Here is a minimal example to reproduce the issue if this code is copied into a file called `example_actor.py`:
```
"""
Run a rabbitmq broker.

Startup workers with
    dramatiq-gevent example_actor --processes 1 --threads 2
or
    dramatiq example_actor --processes 1 --threads 2

Run
    python example_actor.py
observe logs when running workers with dramatiq vs. dramatiq-gevent.

With dramatiq, the TimeLimitExceeded exception is raised in the thread
successfully.
With dramatiq-gevent, the TimeLimitExceeded exception fails to be raised and a
CRITICAL log line is logged.

The behaviour is identical if a redis broker is used instead.
"""

import time

import dramatiq
from dramatiq.brokers.rabbitmq import RabbitmqBroker


[...]

PR #408, opened 2021-07-02
Add support for gevent to TimeLimit middleware

FinnLidbetter:
Fixes Issue #403

- Adds function to check if gevent monkey patching is in use
- Adds gevent to list of extra dependencies
- Makes Time Limit middleware use gevent Timeout objects if gevent is
  being used
- Adds testing for the time limit middleware (for gevent and non-gevent)
- Adds testing script for running pytest with gevent monkey-patching
- Adds gevent pytest command to tox with --cov-append.
- Identifies existing tests that fail with gevent (shutdown notification
  middleware). I suggest that fixing this other issue should be out of
  scope for this PR.

Note that this *doubles* the length of time that it takes tests to run
since tests are run once with gevent and once without. This approach to
testing of running *everything* twice should be considered carefully
before accepting. 
I am open to revising this to only running certain marked tests with gevent,
but running the whole test suite with gevent does not seem inappropriate
to me and it is certainly safer for catching regressions and identifying future 
incompatibilities.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -48,3 +48,4 @@ of those changes to CLEARTYPE SRL.
 | [@denhai](https://github.com/denhai)                  | Hayden Bartlett        |
 | [@rouge8](https://github.com/rouge8)                  | Andy Freeland          |
 | [@thomazthz](https://github.com/thomazthz)            | Thomaz Soares          |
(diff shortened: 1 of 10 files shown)


Bogdanp:
Thanks! This has been rebased and merged.

Issue #419, opened 2021-09-12
PyThreadState_SetAsyncExc c type mismatch


FinnLidbetter:
In the threading.py module the `ctypes.pythonapi.PyThreadState_SetAsyncExc` function is passed a `thread_id` argument as a `ctypes.c_long` object. However, it looks like since python3.7 the type of this argument should be `ctypes.c_ulong`. See [https://docs.python.org/3/c-api/init.html#c.PyThreadState_SetAsyncExc](https://docs.python.org/3/c-api/init.html#c.PyThreadState_SetAsyncExc)

I have not encountered this causing a problem yet, but if I am not mistaken it looks like it could be a bug when using python3.7 and beyond.

I can submit a PR for this to condition the ctype on the python version.

PR #420, opened 2021-09-12
Fix AsyncExc thread_id ctype for python >= 3.7

FinnLidbetter:
Fixes issue #419


FinnLidbetter:
Ah, apparently sometimes c_ulong(-1) will be 2^32 -1 and the same as c_uint(-1), and others it will be 2^64 - 1 [https://docs.python.org/3/library/ctypes.html#ctypes.c_uint](https://docs.python.org/3/library/ctypes.html#ctypes.c_uint)

Bogdanp:
Thank you!

PR #429, opened 2021-10-06
Fix: retries counter increased on last try

### LINKED ISSUE #425: `retries` counter does not increment on last retry
# Issues

GitHub issues are for bugs.  If you have questions, please ask them on the [discussion board](https://reddit.com/r/dramatiq).

## Checklist

* [x] What OS are you using?
* [x] What did you do?
* [x] What happened?

## What OS are you using?

OS X 11.5.2

1.11.0

## What did you do?

I'm trying to run a `on_failure` callback in combination with `retry`. As stated in the documentation, the `on_failure` callback will run after every retry. Therefore I added a check in the callback to only run after a certain number of retries:

```python
MAX_RETRIES = 3

@dramatiq.actor
def my_callback(message_data, exception_data):
    retries = message_data["options"]["retries"]
    if retries > MAX_RETRIES:
        print("handling the error")

@dramatiq.actor(max_retries=MAX_RETRIES, on_failure="my_callback")
def my_actor(*args, **kwargs):
    x = 1/0
```

That the `my_callback` function would run exactly once.

## What happened?

`retries` never exceeds `MAX_RETRIES` and therefore the callback doesn't execute the actual handling code.
On the other hand, if I set `retries >= MAX_RETRIES` the callback gets executed *twice*.

## Proposal

I think this would be fixed by moving [this increment](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/middleware/retries.py#L103) before [this if block](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/middleware/retries.py#L97-L101). 

Happy to make a PR if you think that (a) this is in fact an issue and (b) the proposed solution makes sense.

PS. Thanks for the great library ;)

giuppep:
Closes #425 

I added the test to the `test_callbacks.py` file. Let me know if it should be somewhere else.

### DIFF
--- a/dramatiq/middleware/retries.py
+++ b/dramatiq/middleware/retries.py
@@ -104,7 +104,6 @@ class Retries(Middleware):
             message.fail()
             return
 
-
         if isinstance(exception, Retry) and exception.delay is not None:
             delay = exception.delay
         else:


Bogdanp:
Thanks!

Issue #433, opened 2021-10-08
Redis server crash when delayed messages are greater than the LUA stack


ethervoid:
# Issues

GitHub issues are for bugs.  If you have questions, please ask them on the [discussion board](https://reddit.com/r/dramatiq).

## Checklist

* [x] Does your title concisely summarize the problem?
* [ ] Did you include a minimal, reproducible example?
* [x] What OS are you using?
* [x] What version of Dramatiq are you using?
* [x] What did you do?
* [x] What did you expect would happen?
* [x] What happened?

## What OS are you using?

ECS Fargate (probably is Amazon linux)

## What version of Dramatiq are you using?
1.11.0

## What did you do?
We're running the last version of Dramatiq, when we suffered a huge peak of traffic and received information from our Redis provider that our cluster was crashing with this message: `lua stack limit reach when parsing redis.call reply`

So the problem is we reached the limit of the LUA stack for our Redis interpreter, which has an 8000 limit.

The command being executed was:

```
argv[0]: 'EVALSHA'
argv[1]: '56816d9e79234cb13c1451e28e4bde853b6cae2e'
argv[2]: '1'
argv[3]: 'dramatiq'
argv[4]: 'fetch'
argv[5]: '1642399652204'
argv[6]: 'dispatch_events_queue.DQ'
argv[7]: 'f025a5dc-8a51-4ce9-b07c-7759374571cd'
argv[8]: '60000'
argv[9]: '604800000'
argv[10]: '0'
argv[11]: '7992'
argv[12]: '8000'
```

## What did you expect would happen?

No server crash if the number of delayed messages are increasing

## What happened?

The server crashed, the AOF register write commands, and failed to restart again due to not be able to execute the command and crashing again due to the StackOverflow.

I've been checking the Redis code and basically, the server crashes when the command is not able to perform an increase in the LUA stack, like [this example](https://github.com/redis/redis/blob/b874c6f1fcf844d28f84fc2a2ab6b8f6c2458462/src/scripting.c#L172-L175) , but the weird thing is that every script, `EVALSHA` execution, is atomic so the crash could only happen in the script execution.

As you can see in the crashed command, we're getting 7992 messages and performing actions on them but then we need to convert types from Redis to LUA, and then we're growing more than 1 element in the stack, for example [here](https://github.com/redis/redis/blob/b874c6f1fcf844d28f84fc2a2ab6b8f6c2458462/src/scripting.c#L270-L274) we're trying to increase the current stack by 3

[...]

PR #434, opened 2021-10-12
Lua stack and prefetch variables

ethervoid:
We've found that with a great number of delayed messages, using the max LUA stack in the prefetch for delayed messages and multiple workers you can end up having stack overflow in Redis and crashing the LUA script so this PR intends to minimize that situation:

- Adding configurable environment variable for the delayed queue so we can define the maximum amount of delayed messages fetched
- Adding configurable environment variable to define the max LUA stack value
- Set the maximum unpacking value to 1/2 of the max LUA stack

Related to https://github.com/Bogdanp/dramatiq/issues/433


Bogdanp:
Thanks! I've merged the change with rebase.

Re. `tox`, did you run it with a specific environment (eg. `tox -e py38-cpython`)?  When you do that, it should automatically install everything.

PR #439, opened 2021-10-30
use set.discard to remove item

andruskutt:
https://docs.python.org/3.9/library/stdtypes.html#frozenset.discard


Bogdanp:
Thanks!

PR #443, opened 2021-11-05
Document results time-to-live

miohtama:
While this was covered in the API documentation, it was a bit hidden and this patch will bring `result_ttl` to the narrative documentation.


Bogdanp:
Thanks! I made some of my own changes on top and merged the change with rebase.

PR #449, opened 2021-11-11
Stuff SkipMessage and fail skipped messages

### LINKED ISSUE #440: Skipped messages can get inconsistent results stored
If a message is dropped because of the Age Limit middleware, then getting the result of that message will be inconsistent depending on how soon after the message is dropped an attempt to get the result is made.

Not sure what the resolution should be here. I think it should either consistently store a None result, or consistently store an exception that is not `Exception("unknown")`.

## What OS are you using?

Ubuntu 20.04

1.11.0

## What did you do?

Use a broker with the AgeLimit and Results middleware. Have a message hit its age limit and get its result in a blocking get_result call. Get the result for the same message again a second or two after the blocking get_result returns a result.

The result should not change.

## What happened?

On the first get_result, the result is None. 
On the second get_result, retrieving the result raises a ResultFailure exception with exception `Exception("unknown")`

This appears to be because:

1. When the Age Limit expires, the AgeLimit middleware sets the `failed` flag on the MessageProxy before the message is processed https://github.com/Bogdanp/dramatiq/blob/e1f32459fd837c315f298d9b25b81c1a2f2ab339/dramatiq/middleware/age_limit.py#L41-L50
2. In worker.py this causes and emit_after of "process_message" with a result of None https://github.com/Bogdanp/dramatiq/blob/e1f32459fd837c315f298d9b25b81c1a2f2ab339/dramatiq/worker.py#L493
3. The Results middleware's `after_process_message` will then store this None result and no exception https://github.com/Bogdanp/dramatiq/blob/e1f32459fd837c315f298d9b25b81c1a2f2ab339/dramatiq/results/middleware.py#L95
4. After this happens, worker.py still has some post-processing of the message to do. Since the message was marked with failed, post_process_message will do an `emit_after` of "nack" https://github.com/Bogdanp/dramatiq/blob/e1f32459fd837c315f298d9b25b81c1a2f2ab339/dramatiq/worker.py#L347
5. Since there was no exception, the results middleware will then replace the result at the message key in the result store with https://github.com/Bogdanp/dramatiq/blob/e1f32459fd837c315f298d9b25b81c1a2f2ab339/dramatiq/results/middleware.py#L105-L109

## Sample Code

```
import dramatiq
import time

from dramatiq.brokers.rabbitmq import RabbitmqBroker
from dramatiq.results.backends import RedisBackend
from dramatiq.results import Results

result_backend = RedisBackend()
broker = RabbitmqBroker()
broker.add_middleware(Results(backend=result_backend))
dramatiq.set_broker(broker)

@dramatiq.actor(store_results=True, max_age=1)
def sample_actor():
  return 42

if __name__ == "__main__":
    message = sample_actor.send()
    result_1 = message.get_result(block=True, timeout=2000)
    print(result_1)
    time.sleep(2)
    # An Exception("unknown") will be raised here.
    result_2 = message.get_result()
    print(result_2)
```

FinnLidbetter:
Fixes #440

Stuff SkipMessage exceptions into the message proxy to ensure that
the result is consistent when used in conjunction with the Results
middleware.
Additionally, move the message.fail() call into the SkipMessage
handling to ensure that skipped messages are nack'd rather than
ack'd.

Note that if failing the message is left to the middleware,
then unintentionally omitting message.fail() can lead to no result
ever being stored by the results middleware for the message.
If it should be left to the middleware to decide to fail the skipped
message then it would be more sensible to have fail as the default.
This could be achieved by making the SkipMessage exception take an
optional "fail_message" parameter that defaults to True, and have
the message conditionally fail() inside the worker SkipMessage
exception handling. Custom middleware could then raise SkipMessage
with fail_message=False to override this behaviour and lead to the
message being skipped and ack'd rather than skipped and nack'd.

### DIFF
--- a/dramatiq/middleware/age_limit.py
+++ b/dramatiq/middleware/age_limit.py
@@ -46,4 +46,5 @@ class AgeLimit(Middleware):
 
         if current_millis() - message.message_timestamp >= max_age:
             self.logger.warning("Message %r has exceeded its age limit.", message.message_id)
+            message.fail()
             raise SkipMessage("Message age limit exceeded")
--- a/dramatiq/middleware/age_limit.py
+++ b/dramatiq/middleware/age_limit.py
@@ -17,7 +17,7 @@
 
 from ..common import current_millis
 from ..logging import get_logger
-from .middleware import Middleware
+from .middleware import Middleware, SkipMessage
 
 
 class AgeLimit(Middleware):
@@ -46,5 +46,4 @@ class AgeLimit(Middleware):
 
         if current_millis() - message.message_timestamp >= max_age:
             self.logger.warning("Message %r has exceeded its age limit.", message.message_id)
-            message.fail()
-            return
+            raise SkipMessage("Message age limit exceeded")
--- a/dramatiq/worker.py
+++ b/dramatiq/worker.py
@@ -496,7 +496,8 @@ class _WorkerThread(Thread):
             self.broker.emit_after("process_message", message, result=res)
 
         except SkipMessage as e:
-            message.stuff_exception(e)
+            if message.failed:
+                message.stuff_exception(e)
             self.logger.warning("Message %s was skipped.", message)
             self.broker.emit_after("skip_message", message)
 
--- a/dramatiq/worker.py
+++ b/dramatiq/worker.py
@@ -466,7 +466,10 @@ class _WorkerThread(Thread):
 
     def process_message(self, message):
         """Process a message pulled off of the work queue then push it
-        back to its associated consumer for post processing.
+        back to its associated consumer for post processing. Stuff any SkipMessage
+        exception or BaseException into the message [proxy] so that it may be used
+        by the stub broker to provide a nicer testing experience. Also used by the
+        results middleware to pass exceptions into results.
 
         Parameters:
           message(MessageProxy)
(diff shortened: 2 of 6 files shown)


dramatiq/worker.py
```diff
@@ -492,15 +495,13 @@ def process_message(self, message):
 
             self.broker.emit_after("process_message", message, result=res)
 
-        except SkipMessage:
+        except SkipMessage as e:
+            message.fail()
```
Bogdanp, line 499: `SkipMessage`s shouldn't be marked as having failed. The idea behind `SkipMessage` is to let users ack messages that they don't intend to process. Nacking means messages get put on the DLQ, which is semantically the wrong thing here.

dramatiq/middleware/age_limit.py
```diff
@@ -46,5 +46,4 @@ def before_process_message(self, broker, message):
 
         if current_millis() - message.message_timestamp >= max_age:
             self.logger.warning("Message %r has exceeded its age limit.", message.message_id)
-            message.fail()
```
Bogdanp, line 49: I think the right change here is to retain the `message.fail()` line _and_ `raise SkipMessage`.

Bogdanp (changes requested):

dramatiq/worker.py
FinnLidbetter: Ok that makes sense. Thanks for the feedback. But what would you expect the stored result to be if a user implements a middleware that skips messages, but does not fail them? Would it be `None` or a `SkipMessage` exception? If it should be a `SkipMessage` exception then I can work this in by adding an `after_skip_message` implementation to the results middleware that is similar to its `after_nack` implementation, but checks for a non-None `message._exception` (rather than `message.failed`) to store the exception.

Bogdanp: I think the result of a skipped message should probably be `None`.

Adding `after_skip_message` to the `Results` middleware conflicts with the `after_nack` hook (well, it doesn't exactly "conflict", but it's confusing): the `AgeLimit` middleware both fails the message and raises a Skip, so both hooks run, and it happens to be that `after_skip` runs first, then `after_nack`, so things work out OK in that case. 

More on `None` vs exception: neither option is ideal, but my feeling is it's more natural for a skipped message's result to be `None` than it is for it to be an exception. Both are equally bad in code that doesn't expect them, so we should document this interaction between `AgeLimit` and `Results`. In code that expects actors to potentially skip messages, dealing with `None` is simpler.

FinnLidbetter: Ok that makes sense. I agree that neither option is ideal. I have put up another commit addressing this.

PR #450, opened 2021-11-13
Protect workers from early main process SIGHUP

### LINKED ISSUE #441: SIGHUP on main process startup can result in orphaned workers
If a SIGHUP is received on the main process after the worker processes have started up, but before the signal handler is defined, then a SIGHUP (or SIGTERM, I think) can cause the worker processes to be left behind, still running, but without a main process monitoring them.

This is more of an edge case that is not a big deal to me, but I thought I would flag it anyway. It came about from testing a custom middleware similar to the example MaxTasksPerChild middleware that signals the main process with SIGHUP. This shouldn't be an issue for anyone writing their own middleware that signals the main process, since I think it would be impossible to have a middleware trigger a SIGHUP before the signal handler is defined on the main process. But it could be an issue for anyone with an external process that SIGHUPs dramatiq based on some condition---although perhaps the onus should be on them to not send the Dramatiq main process a SIGHUP too close to a startup/restart?

## What OS are you using?

Ubuntu 20.04

1.11.0

## What did you do?

1. Start a Dramatiq main process
2. Send several SIGHUP signals to the Dramatiq main process in quick succession (can be done manually with a terminal---it doesn't need to be so fast that the SIGHUPs need to be triggered programmatically in a tight loop)

Dramatiq would either restart gracefully, ignore signals, or terminate gracefully.

## What happened?

The Dramatiq main process exited and new Dramatiq workers were left behind without a main process managing them. These workers were still able to pick up work from the queues.

FinnLidbetter:
Fixes #441

Note that the added testcase does not cover the issue described
in #441. The added test is a regression test only.
I am struggling to get a reliable test working that consistently
fails on master. The validation for the bugfix for #441 is manual
only.
Let me know if you would like to block merging this until an
automated test is added.

### DIFF
--- a/dramatiq/cli.py
+++ b/dramatiq/cli.py
@@ -386,6 +386,13 @@ def worker_process(args, worker_id, logging_pipe, canteen, event):
     if hasattr(signal, "SIGBREAK"):
         signal.signal(signal.SIGBREAK, termhandler)
 
+    # Unblock the blocked signals inherited from the parent process.
+    if hasattr(signal, "pthread_sigmask"):
+        signal.pthread_sigmask(
+            signal.SIG_UNBLOCK,
+            {signal.SIGINT, signal.SIGTERM, signal.SIGHUP},
+        )
+
     running = True
     while running:
         time.sleep(1)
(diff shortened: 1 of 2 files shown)


Bogdanp:
Thanks! This is now merged with rebase. I don't think we need an automated test for the fix.

PR #458, opened 2021-12-02
feat(rabbitmq): rely on per-message ttl for delayed tasks

timdrijvers:
### Proof-of-concept. 

Rabbitmq 3.8.25 (at AWS, using Amazon MQ RabbitMQ broker) has a default, not configurable, `consumer_timeout` of 30 minutes. Handling of delayed messages in Dramatiq requires the worker to prefetch messages and queue these internally. If a message is queued beyond the consumer timeout deadline, the connection raises an exception and is closed. This is causing problems when the worker is also processing tasks at the same time, when the task is finished, a new connection is available where the message tag is unknown. Also causing tasks to be redelivered, causing tasks to be executed twice.

### Solution
The proposed solution relies on a per-message ttl, where dramatiq doesnt need a separate worker to prefetch and queue delayed tasks internally.

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -287,6 +287,7 @@ class RabbitmqBroker(Broker):
 
     def _declare_dq_queue(self, queue_name):
         arguments = self._build_queue_arguments(queue_name)
+        arguments["x-dead-letter-routing-key"] = queue_name  # Route messages to regular queue
         return self.channel.queue_declare(queue=dq_name(queue_name), durable=True, arguments=arguments)
 
     def _declare_xq_queue(self, queue_name):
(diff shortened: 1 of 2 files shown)


timdrijvers:
Seems that this solution direction doesn't work: https://www.rabbitmq.com/ttl.html#per-message-ttl

> When setting per-message TTL expired messages can queue up behind non-expired ones until the latter are consumed or expired. Hence resources used by such expired messages will not be freed, and they will be counted in queue statistics (e.g. the number of messages in the queue).

Bogdanp:
I think there's no good workaround here. Vanilla RMQ doesn't have good support for delayed messages, hence the hacky way dramatiq implements them, but the timeout defeats that hack, and, as you've found, the TTL approach has its downsides. 

According to [this thread](https://forums.aws.amazon.com/thread.jspa?messageID=994285&tstart=0), version 3.8.11 doesn't have the timeout, so probably the best approach for you guys here would be to roll back to that version until they add support for configuring the timeout in a future version.

PR #471, opened 2022-01-29
Adds closed attribute to StreamablePipe

nathanielobrown:
Some libraries (in my case [playwright](https://github.com/microsoft/playwright-python)) try to access `sys.<stdout|stderr>.closed` which raises an error if run in a Dramatiq actor. This fixes that.

I haven't written any tests because this is such a small change and I do not see any existing tests for `StreamablePipe`. However, let me know if you'd like me to add one.

P.S. I'm really loving Dramatiq! Thank you for all your work


Bogdanp:
Thanks!

PR #473, opened 2022-01-31
add GeventRabbitmqBroker

horpto:
Under gevent dramatiq is opening a connection at each query (with message).


horpto:
@Bogdanp if I'd remove GeventRabbitmqBroker itself and keep other changes as refactoring can it be merged?

Bogdanp:
Thanks, but I don't think the additional hooks are needed since someone can extend the class and override the channel and connection properties instead.

PR #481, opened 2022-03-14
fix: race condition on join() in the redis broker.

### LINKED ISSUE #480: Race condition in redis broker `join`
# Issues

GitHub issues are for bugs.  If you have questions, please ask them on the [discussion board](https://reddit.com/r/dramatiq).

## Checklist

* [x] What OS are you using?
* [x] What did you do?
* [x] What happened?

## What OS are you using?

windows 10 / ubuntu 20.04

1.12.0 and 1.12.3

## What did you do?

A test in an automated pipeline tests that an actor enqueued with a delay is executed after the `join` of the broker returns.

The actor is correctly run

## What happened?

Sometimes the actor does not run, because of a race condition.

Following it's an example code can reproduces the race condition:
```py
import os
import dramatiq
import pytest
from dramatiq.brokers.redis import RedisBroker
from dramatiq.middleware import AgeLimit, Retries, ShutdownNotifications
import pathlib
import logging

logging.basicConfig()
logging.getLogger("dramatiq").setLevel(logging.DEBUG)

_MIDDLEWARES = [AgeLimit(), ShutdownNotifications(), Retries(max_retries=3)]
redis_broker = RedisBroker(
    host="localhost",
    middleware=_MIDDLEWARES,
    namespace=f"ns-{os.getpid()}",
    password="redis-passwd",
)
dramatiq.set_broker(redis_broker)

@dramatiq.actor
def bar(s):
    pathlib.Path(f"file{s}.canary").touch(exist_ok=False)
    print(s)

@pytest.fixture(scope="module")
def wM():
    w = dramatiq.Worker(redis_broker)
    w.start()
    yield w
    w.stop()

N = 500
@pytest.fixture(scope="module")
def clean():
    def do_clean():
        for n in range(N):
            pathlib.Path(f"file{n}.canary").unlink(missing_ok=True)
    do_clean()
    yield
    do_clean()

@pytest.mark.parametrize("n", range(N))
def test(n, wM, clean):
    bar.send_with_options(kwargs={"s": str(n)}, delay=10)
    wM.broker.join(queue_name="default")
    wM.join()
    assert pathlib.Path(f"file{n}.canary").exists()
```
This can be run with pytest using `pytest -x`. Note that it can be quite rare on some environments, so a single run may not be sufficient. parallelizing with `-n` seems to help reproducing this, but it's not required to reporduce.

To better illustrate investigate the issue I've added some logs in the redis broker join operation:
```diff
diff --git a/dramatiq/brokers/redis.py b/dramatiq/brokers/redis.py   
index 0f37313..838595d 100644
--- a/dramatiq/brokers/redis.py
+++ b/dramatiq/brokers/redis.py
@@ -233,7 +233,10 @@ class RedisBroker(Broker):
 
             size = 0
             for name in (queue_name, dq_name(queue_name)):
-                size += self.do_qsize(name)
+                current = self.do_qsize(name)
+                self.logger.debug('queue %s size %s', name, current)
+                size += current
+            self.logger.debug('join total size %s', size)
 
             if size == 0:
                 return
```

with this change in place a failure logs the following:
```
===================================================================== FAILURES ======================================================================
_____________________________________________________________________ test[300] _____________________________________________________________________
[gw1] win32 -- Python 3.8.12 path\to\python.exe

n = 300, wM = <dramatiq.worker.Worker object at 0x0000012105930070>, clean = None

    @pytest.mark.parametrize("n", range(500))
    def test(n, wM, clean):
        bar.send_with_options(kwargs={"s": str(n)}, delay=10)
        wM.broker.join(queue_name="default")
        wM.join()
>       assert pathlib.Path(f"file{n}.canary").exists()
E       AssertionError: assert False
E        +  where False = <bound method Path.exists of WindowsPath('file300.canary')>()
E        +    where <bound method Path.exists of WindowsPath('file300.canary')> = WindowsPath('file300.canary').exists
E        +      where WindowsPath('file300.canary') = <class 'pathlib.Path'>('file300.canary')
E        +        where <class 'pathlib.Path'> = pathlib.Path

tests\a_test.py:70: AssertionError
----------------------------------------------------------------- Captured log call -----------------------------------------------------------------
DEBUG    dramatiq.broker.RedisBroker:redis.py:184 Enqueueing message '0619fc48-c787-4125-999e-8d439321e0ed' on queue 'default.DQ'.
DEBUG    dramatiq.broker.RedisBroker:redis.py:238 queue default size 0
DEBUG    dramatiq.broker.RedisBroker:redis.py:238 queue default.DQ size 1
DEBUG    dramatiq.broker.RedisBroker:redis.py:240 join total size 1
DEBUG    dramatiq.worker.ConsumerThread(default.DQ):worker.py:320 Pushing message '0619fc48-c787-4125-999e-8d439321e0ed' onto delay queue.
DEBUG    dramatiq.broker.RedisBroker:redis.py:184 Enqueueing message '0619fc48-c787-4125-999e-8d439321e0ed' on queue 'default'.
DEBUG    dramatiq.broker.RedisBroker:redis.py:238 queue default size 0
DEBUG    dramatiq.worker.ConsumerThread(default.DQ):worker.py:350 Acknowledging message '0619fc48-c787-4125-999e-8d439321e0ed'.
DEBUG    dramatiq.broker.RedisBroker:redis.py:238 queue default.DQ size 0
DEBUG    dramatiq.broker.RedisBroker:redis.py:240 join total size 0
============================================================== short test summary info ==============================================================
```

The race condition seems to be the following:
- the message delay has elapsed and processing of the message is in progress in the worker thread
- in the main thread the default queue size gets checked and returns 0
- the worker thread pushes the message to the default queue, then acks the message in the delay queue
- in the main thread the delay queue size gets checked and returns 0
- the join returns, but a message is present in the default queue.

The fix seems to be very easy, and it's just to swap in which order the queues are checked.
```diff
diff --git a/dramatiq/brokers/redis.py b/dramatiq/brokers/redis.py
index 0f37313..ce4aa98 100644
--- a/dramatiq/brokers/redis.py
+++ b/dramatiq/brokers/redis.py
@@ -232,7 +232,7 @@ class RedisBroker(Broker):
                 raise QueueJoinTimeout(queue_name)

             size = 0
-            for name in (queue_name, dq_name(queue_name)):
+            for name in (dq_name(queue_name), queue_name):
                 size += self.do_qsize(name)

             if size == 0:
```

This this change in place I've run the test above 20 times without any failure.

I'm opening a PR with the suggested fix

CaselIT:
This fixes #480 

I'm not sure how to test this, since the race condition is not easy to trigger. @Bogdanp do you have any suggestions?

Also looking at the stub broker it seems that it may have a similar race condition, but I've never triggered a similar issue with it, so I'm not sure if the change (swapping the order or the delay queue and normal one) is needed
https://github.com/Bogdanp/dramatiq/blob/2d0ca2519a0289946337b308e450a915cf5fd869/dramatiq/brokers/stub.py#L153-L164

### DIFF
--- a/dramatiq/brokers/redis.py
+++ b/dramatiq/brokers/redis.py
@@ -231,11 +231,7 @@ class RedisBroker(Broker):
             if deadline and time.monotonic() >= deadline:
                 raise QueueJoinTimeout(queue_name)
 
-            size = 0
-            # The delay queue needs to be checked before the normal queue.
-            # See https://github.com/Bogdanp/dramatiq/issues/480 for details.
-            for name in (dq_name(queue_name), queue_name):
-                size += self.do_qsize(name)
+            size = self.do_qsize(queue_name)
 
             if size == 0:
                 return
--- a/dramatiq/brokers/redis.py
+++ b/dramatiq/brokers/redis.py
@@ -232,7 +232,9 @@ class RedisBroker(Broker):
                 raise QueueJoinTimeout(queue_name)
 
             size = 0
-            for name in (queue_name, dq_name(queue_name)):
+            # The delay queue needs to be checked before the normal queue.
+            # See https://github.com/Bogdanp/dramatiq/issues/480 for details.
+            for name in (dq_name(queue_name), queue_name):
                 size += self.do_qsize(name)
 
             if size == 0:
--- a/dramatiq/brokers/redis/dispatch.lua
+++ b/dramatiq/brokers/redis/dispatch.lua
@@ -229,8 +229,17 @@ elseif command == "purge" then
     redis.call("del", queue_full_name, queue_acks, queue_messages, xqueue_full_name, xqueue_messages)
 
 
--- Used in tests to determine the size of the queue.
+-- Used in tests to determine the size of the queue and its connected delay queue.
 elseif command == "qsize" then
-    return redis.call("hlen", queue_messages) + redis.call("scard", queue_acks)
+    local q_aks = acks .. "." .. queue_canonical_name
+    local q_fn = namespace .. ":" .. queue_canonical_name
+    local q_msgs = q_fn .. ".msgs"
+    local dq_q_aks = q_aks .. ".DQ"
+    local dq_q_msgs = q_fn .. ".DQ" .. ".msgs"
+
+    return redis.call("hlen", dq_q_msgs) +
+           redis.call("scard", dq_q_aks) +
+           redis.call("hlen", q_msgs) +
+           redis.call("scard", q_aks)
 
 end
--- a/tests/test_redis.py
+++ b/tests/test_redis.py
@@ -496,40 +496,28 @@ def test_redis_join_race_condition(redis_broker, redis_worker):
     """
     test for issue https://github.com/Bogdanp/dramatiq/issues/480
 
-    this test was triggered a race condition when ``qsize`` returned only the size of the
-    requested queue, not also of the delay queue.
+    this test verifies that do_qsize counts all messages in the queue and delay queue at the same
+    time, removing the race condition mentioned in the issue
     """
-    old_qsize = redis_broker.do_qsize
-    events = defaultdict(threading.Event)
-
-    class M(dramatiq.Middleware):
-        def before_enqueue(self, broker, message, delay):
-            if not delay:
-                events["default"].set()
-                events["default_done"].wait(2)
-
-        def after_ack(self, broker, message):
-            if message.options.get("eta"):
-                events[dq_name("default")].set()
-
-    def qsize(qn):
-        events[qn].wait(2)
-        events[qn].clear()
-        res = old_qsize(qn)
-        events[f"{qn}_done"].set()
-        return res
-
-    redis_broker.add_middleware(M())
-    redis_broker.do_qsize = qsize
     called = False
+    size = []
 
     @dramatiq.actor
     def go():
+        go_again.send_with_options(delay=50)
+        size.append(redis_broker.do_qsize('default'))  # go ack + go msg + go_again msg
+        size.append(redis_broker.do_qsize(dq_name('default')))  # does the same
+
+    @dramatiq.actor
+    def go_again():
         nonlocal called
         called = True
+        size.append(redis_broker.do_qsize('default'))  # go_again msg + go_again ack
+        size.append(redis_broker.do_qsize(dq_name('default')))  # does the same
 
-    go.send_with_options(delay=50)
+    go.send()
     redis_broker.join("default")
     redis_worker.join()
 
     assert called
+    assert size == [3, 3, 2, 2]
--- a/tests/test_redis.py
+++ b/tests/test_redis.py
@@ -490,7 +490,7 @@ def test_redis_consumer_nack_does_not_raise_on_missing_id(redis_worker):
     consumer.nack(message)
 
 
-def test_redis_join_race_condition(redis_broker, redis_worker):
+def test_redis_join_race_condition(redis_broker):
     """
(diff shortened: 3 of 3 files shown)


Bogdanp:
I think this likely doesn't get rid of the race. A better fix would be to change [qsize](https://github.com/Bogdanp/dramatiq/blob/2d0ca2519a0289946337b308e450a915cf5fd869/dramatiq/brokers/redis/dispatch.lua#L234) so that it returns the size of the queue _and_ its associated delay queue. Then we can replace the `for` loop with `size = self.do_qsize(queue_name)`. Let me know if you'd like to do that. Otherwise, I'll do it when I have some free time.

CaselIT:
> I think this likely doesn't get rid of the race

yes, the race is still there, but instead of not counting the message, it would be double counted (the delayed message in the delay queue and the normal message in the normal queue) fixing the issue for the `join()` prospective.

That said I agree that's probably better to update qsize, since that would most likely get rid of the rage condition altogether.

Thanks also for the test suggestion, I'll try working it and update the PR

CaselIT:
I've added a test for that specific regression, but I think we may as well improve qsize

CaselIT:
updated. the test now does not make much sense, but it continues to pass

Bogdanp:
Thanks! I think we might as well drop the test at this point since that kind of ordering is no longer possible (and we're unlikely to ever regress).

CaselIT:
Maybe I can change it to check that qsize returns all the messages in the queue & delay queue

CaselIT:
btw @Bogdanp it seems that the ci on windows should be updated, since windows 2016 is being retired

CaselIT:
> Thanks! I think we might as well drop the test at this point since that kind of ordering is no longer possible (and we're unlikely to ever regress).

Updated the test. I think it's ready for a review

CaselIT:
> btw @Bogdanp it seems that the ci on windows should be updated, since windows 2016 is being retired

doing a simple update breaks the libmemcached compile: see the action run on this commit https://github.com/CaselIT/dramatiq/commit/39285f3f0a054a8ef5cdb16792e53eb1f142f86c

Bogdanp:
> doing a simple update breaks the libmemcached compile: see the action run on this commit https://github.com/CaselIT/dramatiq/commit/39285f3f0a054a8ef5cdb16792e53eb1f142f86c

Yes, I'm afraid that's going to be stuck until someone can update https://github.com/ryansm1/libmemcached-win/tree/yshurik-win to work w/ 2019.

CaselIT:
> > doing a simple update breaks the libmemcached compile: see the action run on this commit https://github.com/CaselIT/dramatiq/commit/39285f3f0a054a8ef5cdb16792e53eb1f142f86c
> 
> Yes, I'm afraid that's going to be stuck until someone can update https://github.com/ryansm1/libmemcached-win/tree/yshurik-win to work w/ 2019.

I guess an option is to disable mamcached test on windows. In it doesn't compile chances are that people are not going to use it. I can open a PR of that's an acceptable solution

brianseodd:
@CaselIT would it be safe to install build the module from 1.13.0 to absorb this change? I'm getting a lot of issues with dramatiq that I suspect are resolved by this PR

CaselIT:
I don't think this patch is released on pypi yet

brianseodd:
@CaselIT Yeah I was planning on just building it from master offline if you don't think there is anything stopping this from being viable. I'm getting something related to this and getting errors the below and it seems to be related to when the queue is completely exhausted or something. 

```
ResponseError: Error running script (call to f_1b0bca8bf4560c1f4a84ff992da836e49515f830): @user_script:128: @user_script: 128: Unknown Redis command called from Lua script ```

PR #482, opened 2022-03-15
typing: improved mypy support

### LINKED ISSUE #305: Typing support for the public API

staticdev:
* py.typed mark as required by [PEP-561](https://peps.python.org/pep-0561/)
* add mypy.ini config
* fix JSONEncoder error handling
* some mypy error fixes
* some isort error fixes

Closes #305

### DIFF
--- a/dramatiq/broker.py
+++ b/dramatiq/broker.py
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from typing import Optional
+from typing import cast
 
 from .errors import ActorNotFound
 from .logging import get_logger
@@ -25,7 +25,7 @@ from .middleware import MiddlewareError, default_middleware
 global_broker = None
 
 
-def get_broker() -> Optional["Broker"]:
+def get_broker() -> "Broker":
     """Get the global broker instance.  If no global broker is set,
     this initializes a RabbitmqBroker and returns it.
 
@@ -43,6 +43,7 @@ def get_broker() -> Optional["Broker"]:
             connection_attempts=5,
             blocked_connection_timeout=30,
         ))
+    global_broker = cast("Broker", global_broker)
     return global_broker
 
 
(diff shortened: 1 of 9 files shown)


staticdev:
@Bogdanp there is still a lot of typing missing, but this PR will remove some complaints from mypy of when you import dramatiq.

dramatiq/broker.py
```diff
@@ -23,7 +25,7 @@
 global_broker = None
 
 
-def get_broker() -> "Broker":
```
CaselIT, line 26: I believe this cannot return `None`

dramatiq/results/backends/stub.py
```diff
@@ -29,7 +30,7 @@ class StubBackend(ResultBackend):
         result data.  Defaults to :class:`.JSONEncoder`.
     """
 
-    results = {}
+    results: Dict[str, Tuple[Optional[str], Optional[str]]] = {}
```
CaselIT, line 33: the second value of the tuple is a float

CaselIT (changes requested):

dramatiq/results/backends/stub.py
staticdev: Thanks for spotting that. Fixed!

dramatiq/broker.py
staticdev: @CaselIT I think I found a better way out.. if we cast the type of global_broker to "Broker" it now stop type-checking errors and enables proper evaluation of the global variable.

staticdev: Fixed also.

staticdev:
LGTM? @CaselIT @Bogdanp

CaselIT (approved):

staticdev:
@Bogdanp could we get a new release with latests commits?

Issue #491, opened 2022-06-08
Use SCAN or the like instead of KEYS


Stealthmate:
## Checklist

* [x] Does your title concisely summarize the problem?
* [x] Did you include a minimal, reproducible example?
* [x] What OS are you using?
* [x] What version of Dramatiq are you using?
* [x] What did you do? Deploy to production
* [x] What did you expect would happen?
* [x] What happened?

## What OS are you using?

`python:3.9-alpine`

## What version of Dramatiq are you using?

1.12.3

## What did you do?

Deploy to production

## What did you expect would happen?

Nothing

## What happened?

My Redis provider blocks `KEYS` because it's slow (as described in the [redis docs](https://redis.io/commands/keys/)), so I got an error.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/dramatiq/brokers/redis.py", line 346, in __next__
    data = self.message_cache.pop(0)
IndexError: pop from empty list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/dramatiq/worker.py", line 264, in run
    for message in self.consumer:
  File "/usr/local/lib/python3.9/site-packages/dramatiq/brokers/redis.py", line 365, in __next__
    self.message_cache = messages = self.broker.do_fetch(
  File "/usr/local/lib/python3.9/site-packages/dramatiq/brokers/redis.py", line 283, in do_dispatch
    return dispatch(args=args, keys=keys)
  File "/usr/local/lib/python3.9/site-packages/redis/commands/core.py", line 5704, in __call__
    return client.evalsha(self.sha, len(keys), *args)
  File "/usr/local/lib/python3.9/site-packages/redis/commands/core.py", line 5089, in evalsha
    return self._evalsha("EVALSHA", sha, numkeys, *keys_and_args)
  File "/usr/local/lib/python3.9/site-packages/redis/commands/core.py", line 5073, in _evalsha
    return self.execute_command(command, sha, numkeys, *keys_and_args)
  File "/usr/local/lib/python3.9/site-packages/redis/client.py", line 1235, in execute_command
    return conn.retry.call_with_retry(
  File "/usr/local/lib/python3.9/site-packages/redis/retry.py", line 46, in call_with_retry
    return do()
  File "/usr/local/lib/python3.9/site-packages/redis/client.py", line 1236, in <lambda>
    lambda: self._send_command_parse_response(
[...]

PR #492, opened 2022-06-10
Use scan instead of keys in redis lua script

Stealthmate:
This is my implementation for dealing with #491. I replaced the `KEYS` call with a `SCAN` loop but then Redis complained that you can't write after a non-deterministic command (which `SCAN` is), so I added the `redis.replicate_commands` call at the top. That being said, `KEYS` should also be non-deterministic so I don't know how/why it worked without `replicate_commands` before. Feel free to correct me if there's a better way to do this.

An idea for a future improvement (if needed) is to allow people to set `COUNT` as described [here](https://redis.io/commands/scan/). I didn't bother with that since `COUNT` is apparently just a hint anyway. Performance __may__ degrade a bit if there are a lot of keys, but should be more or less the same for the most part I think.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -55,3 +55,4 @@ of those changes to CLEARTYPE SRL.
 | [@nathanielobrown](https://github.com/nathanielobrown)| Nathaniel Brown        |
 | [@staticdev](https://github.com/staticdev)            | Thiago                 |
 | [@kurtmckee](https://github.com/kurtmckee)            | Kurt McKee             |
+| [@stealthmate](https://github.com/stealthmate)        | Valeri Haralanov       |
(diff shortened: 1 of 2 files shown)


Stealthmate:
Also, I just got another idea. We should probably rewrite the dead workers loop to process all dead workers in a single `SCAN` iteration. Right now we are scanning the whole store for each worker which may or may not be wildly inefficient.

I can implement this as well later if you want.

Stealthmate:
Turns out `replicate_commands` may or may not be available depending on the redis version, so I added some checks to account for that.

 >We should probably rewrite the dead workers loop to process all dead workers in a single SCAN iteration.

I also implemented this.

Bogdanp:
Thanks for the PR! Ultimately, I don't think the extra complexity warrants the change.

Issue #421, opened 2021-09-15
Warnings emitted for unused result backend


DanCardin:
# Issues

Dramatiq outputs warnings at: https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/results/backends/__init__.py#L22 for both redis and memcached regardless of whether you actually use either of them.

## Checklist

* [x] Does your title concisely summarize the problem?
* [x] What version of Dramatiq are you using?
* [x] What did you do?
* [x] What did you expect would happen?
* [x] What happened?

## What version of Dramatiq are you using?

1.11, although this file hasn't changed since 2018

## What did you do?
I'm only personally using the redis backend, but also should I only attempt to use the StubBackend in tests, these warnings will be emitted.

## What did you expect would happen?

It seems like it'd be preferable to either only emit that warning:
* when an instance is instantiated
* or when the memcached/redis modules are imported

But given that both are preemptively imported, they'll always emit, unless you install all the backend extras.

My personal preference would be to avoid importing any of the optional backends from the `__init__.py`, and doing the import guard from inside the module itself (presumably on the imports for their respective clients).

Then upon attempting to import i.e. `from dramatiq.results.backends.memcached import MemcachedBackend`, one would receive that warning; and upon actually attempting to use it, you'd get the same behavior as before.

## What happened?

Warnings

DanCardin:
My personal preference would be to avoid importing any of the optional backends from the `__init__.py`, and doing the import guard from inside the module itself (presumably on the imports for their respective clients).

Then upon attempting to import i.e. `from dramatiq.results.backends.memcached import MemcachedBackend`, one would receive that warning; and upon actually attempting to use it, you'd get the same behavior as before.

For what it's worth, I'd be happy to submit a PR to this effect ^, if it would be accepted; although i can appreciate that the change in import would be a potentially breaking change and you might want to do something else

Bogdanp:
Addressed by https://github.com/Bogdanp/dramatiq/commit/b8ac60c695c568b4b0daf641a2327a9c9bf1e7f6 .

PR #506, opened 2022-09-27
fix: Avoid warning on missing dependencies when those dependencies are used.

DanCardin:
Defer imports of backend client libraries until the backend is instantiated. This should maintain the existing failure behavior which lazily causes usages of missing backends to fail. However it defers the emission of warnings on unused backends **until** they're instantiated

Addresses https://github.com/Bogdanp/dramatiq/issues/421

This is perhaps somewhat of an suboptimal way of "fixing" the problem, but due to lack of response on the issue, I opted for an option with perhaps the least chance of it breaking existing users, since you expose both RedisBackend/MemcachedBackend throught `__all__` which arguably makes them part of your expected public interface.

I'm happy to adjust to other options if this isn't how you'd ideally fix it.

### DIFF
--- a/dramatiq/results/backends/__init__.py
+++ b/dramatiq/results/backends/__init__.py
@@ -19,20 +19,46 @@ import warnings
 
 from .stub import StubBackend
 
-try:
-    from .memcached import MemcachedBackend
-except ImportError:  # pragma: no cover
-    warnings.warn(
-        "MemcachedBackend is not available.  Run `pip install dramatiq[memcached]` "
-        "to add support for that backend.", ImportWarning,
-    )
-
-try:
-    from .redis import RedisBackend
-except ImportError:  # pragma: no cover
-    warnings.warn(
-        "RedisBackend is not available.  Run `pip install dramatiq[redis]` "
-        "to add support for that backend.", ImportWarning,
-    )
-
 __all__ = ["StubBackend", "MemcachedBackend", "RedisBackend"]
+
+
+def __getattr__(name):
+    module_importer = _module_importers.get(name)
+    if module_importer is None:
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+    return module_importer()
+
+
+def import_memcached():
+    try:
+        from .memcached import MemcachedBackend
+
+        return MemcachedBackend
+    except ModuleNotFoundError:
+        warning = ImportWarning(
+            "MemcachedBackend is not available.  Run `pip install dramatiq[memcached]` "
+            "to add support for that backend.",
+        )
+        warnings.warn(warning)
+        raise
+
+
+def import_redis():
+    try:
+        from .redis import RedisBackend
+
+        return RedisBackend
+    except ModuleNotFoundError:
+        warning = ImportWarning(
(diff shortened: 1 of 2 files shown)


dramatiq/results/backends/__init__.py
```diff
@@ -15,24 +15,8 @@
...
-    from .redis import RedisBackend
-except ImportError:  # pragma: no cover
-    warnings.warn(
-        "RedisBackend is not available.  Run `pip install dramatiq[redis]` "
-        "to add support for that backend.", ImportWarning,
-    )
+from .memcached import MemcachedBackend
+from .redis import RedisBackend
 
 __all__ = ["StubBackend", "MemcachedBackend", "RedisBackend"]
```
Bogdanp, line 22: How about making imports lazy with a module-level `__getattr__`? That way, the changes to `memcached.py` and `redis.py` won't be necessary.  Something like this pseudocode:

```python
import importlib
import warnings

__all__ = ["StubBackend", "MemcachedBackend", "RedisBackend"]

def __getattr__(name):
  modname = {
    "MemcachedBackend": ".memcached",
    "RedisBackend": ".redis",
    "StubBackend": ".stub",
  }.get(name)
  if mod is None:
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
  try:
    mod = importlib.import_module(modname, __name__)
    return getattr(mod, name)
  except ImportError:
    warnings.warn(...)
    raise
    


```

DanCardin: I suppose you only support 3.7+ so that is an option. Fine by me if you're happy with this as a solution.

DanCardin: Although at least due to your error messages, i might still prefer doing the checks non-dynamically; which is what i'll push up momentarily

DanCardin:
```
  File "/Users/danc/github/dramatiq/dramatiq/results/backends/__init__.py", line 35, in import_memcached
    from .memcached import MemcachedBackend
  File "/Users/danc/github/dramatiq/dramatiq/results/backends/memcached.py", line 18, in <module>
    from pylibmc import Client, ClientPool
ModuleNotFoundError: No module named 'pylibmc'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 1075, in _handle_fromlist
  File "/Users/danc/github/dramatiq/dramatiq/results/backends/__init__.py", line 30, in __getattr__
    return module_importer()
  File "/Users/danc/github/dramatiq/dramatiq/results/backends/__init__.py", line 44, in import_memcached
    raise warning
ImportWarning: MemcachedBackend is not available.  Run `pip install dramatiq[memcached]` to add support for that backend.
```
The output here is slightly different, in terms of the exception raised, although the warning is the same. But I figured because of exception chaining, you still see the error that gets raised today (i.e. the `ModuleNotFoundError`), in addition to the suggestion.

Bogdanp:
I think we'd want to re-raise the original import error in case anybody is relying on getting an import error for conditional purposes.

DanCardin:
Alrighty, updated per that

Bogdanp:
Thanks!

PR #512, opened 2022-11-25
Annotate types for Actor and Message

orsinium:
1. Fix existing type violations reported by mypy.
2. Annotate types for `Actor`, `actor`, and `Message`
3. Use TypeVar with Generic to propagate the return type of the function. For example, if you wrap a function that returns `str`, then mypy will know that calling `Actor` instance also returns `str`.

Something that can be improved later:

1. `Message` is a namedtuple, and so cannot inherit from Generic. It can be solved by using a dataclass instead.
2. Types for arguments of the function can be typed with [Parameter Specification Variables](https://peps.python.org/pep-0612/), but for that, you'll need [typing-extensions](https://pypi.org/project/typing-extensions/) the good news is that it can be static-only dependency by putting it inside of `if TYPE_CHECKING`. But I didn't want to complicate this PR too much.

### DIFF
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -17,16 +17,19 @@
 
 import re
 import time
+from typing import Any, Callable, Dict, Generic, Optional, TypeVar, Union, overload
 
-from .broker import get_broker
+from .broker import Broker, get_broker
 from .logging import get_logger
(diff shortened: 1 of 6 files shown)


Bogdanp:
Thanks!

PR #513, opened 2022-11-26
Make Message Generic of result

orsinium:
1. Turn Message from namedtuple into dataclass
2. Make Message a Generic of the expected result. Now, if you do `actor.send().get_result()`, mypy will correctly infer the result type from the actor type annotation.

### DIFF
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -63,7 +63,7 @@ class Actor(Generic[R]):
         self.options = options
         self.broker.declare_actor(self)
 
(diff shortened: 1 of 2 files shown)


Bogdanp:
Thanks!

anton-petrov:
It's a good idea to transform result message to generic. But, perhaps, because of these changes, part of the tests broke and need to be fixed.

orsinium:
Huh, there is no CI. Alrighty, I'll have a look 👌🏿

PR #516, opened 2022-12-08
Fix tests broken by moving Message to dataclass

orsinium:
The changes in #513 broke tests a bit that relied on Message being a namedtuple, see comment https://github.com/Bogdanp/dramatiq/pull/513#issuecomment-1342164116 by @anton-petrov.

1. Add `pipe_target` into `Message.options` inside of `pipeline` in the reverse order, so that the message dump inside `pipe_target` recursively contains `pipe_target` of the next message.
2. Add `Message.__lt__`, so that messages can be sorted. The sorting order is the same as it was for namedtuple.
3. Make sure `Message.args` is a tuple when restoring from a dump.

### DIFF
--- a/dramatiq/composition.py
+++ b/dramatiq/composition.py
@@ -52,9 +52,8 @@ class pipeline:
             else:
                 messages.append(child.copy())
 
-        rev = messages[::-1]
-        for message, prev_message in zip(rev, rev[1:]):
-            prev_message.options["pipe_target"] = message.asdict()
+        for message, next_message in zip(messages, messages[1:]):
+            message.options["pipe_target"] = next_message.asdict()
 
     def __len__(self):
         """Returns the length of the pipeline.
--- a/dramatiq/composition.py
+++ b/dramatiq/composition.py
@@ -14,13 +14,18 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
 
 import time
+from typing import TYPE_CHECKING, Iterable
 from uuid import uuid4
 
 from .broker import get_broker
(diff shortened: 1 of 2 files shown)


dramatiq/composition.py
```diff
@@ -45,8 +52,9 @@ def __init__(self, children, *, broker=None):
             else:
                 messages.append(child.copy())
 
-        for message, next_message in zip(messages, messages[1:]):
-            message.options["pipe_target"] = next_message.asdict()
+        rev = messages[::-1]
```
Bogdanp, line 55: Is this change necessary?

orsinium: Yes, it is. `asdict` returns a copy of the message content. If you set `pipe_target` going from first to last, the message inside of it won't have its own `pipe_target`. You can try it yourself: remove this line and run the tests, there is one test (I don't remember the name) that reproduces it.

Bogdanp: I see, thanks! So that means the earlier change to `asdict` was technically a change in behaviour. I prefer this new behaviour, but I'll have to think about whether or not it might cause issues downstream.

orsinium: It happens because dataclasses.asdict creates a copy of collections:

image

I think we can preserve the old behavior if we explicitly put the real value of `options` instead of the copy on top of the copied result.

Bogdanp: > I think we can preserve the old behavior if we explicitly put the real value of options instead of the copy on top of the copied result.

Yes, I think we should do that.

orsinium: Done. I made `Message.asdict` not to create copies and removed reverse loop from `pipeline.__init__`

Bogdanp:
Thanks!

PR #523, opened 2023-01-08
[DQ-520] Fixes port usage problem when thousands of threads are created

ujnamss:
[DQ-520] Fixes port usage problem when thousands of threads are created


Bogdanp:
Thanks, but I don't consider the existing broker to be broken, and a "v2" implementation isn't the right solution at the moment anyway.

Issue #531, opened 2023-02-24
py.typed file is not present in built distributions on pypi


sirosen:
# Issue

At `$WORK`, we found that `mypy` type checking didn't work with `dramatiq`, which surprised us because of the presence of the `py.typed` in the repo here.

Investigating, the setuptools config seems fine, and I'm not able to reproduce a build which omits this file using a modern version of setuptools or using pypa's `build` to run `python -m build .`.

## What version of Dramatiq are you using?

1.14.0

## What did you do?

Install `dramatiq` and `mypy` into a virtualenv, write a file which runs `reveal_type(dramatiq.set_encoder)`, and run `mypy` on that file.

```
# foo.py
import dramatiq

reveal_type(dramatiq.set_encoder)
```

## What did you expect would happen?

```
$ mypy foo.py
foo.py:3: note: Revealed type is "def (encoder: dramatiq.encoder.Encoder)"
```

## What happened?

```
$ mypy foo.py
foo.py:3: note: Revealed type is "Any"
```

## Fixing this

It's not clear how to submit a code change which could fix this because it's not clear what the publishing workflow for `dramatiq` is. I believe this can be handled with some codified workflow which uses `python -m build`. e.g. I've put this into `tox` in many of my projects as `tox r -e build`.

I'm happy to contribute a fix which adds some canonical build workflow, but don't know that it's necessary or desirable.

sirosen:
Oh, also, I downloaded the sdist and wheel from `pypi` for 1.14.0 and ran `tar -tzf` and `unzip -l` on them, and did not see `py.typed` in the file list.

Bogdanp:
Thank! I've just published 1.14.1, which should fix this problem. 

Re. the workflow, currently it's a manual process where I 

1. run `bumpversion` with the relevant modifier (`patch` or `minor`)
2. run `python setup.py sdist bdist_wheel`
3. run `twine upload -s dist/dramatiq-$version`

I'd happily accept a PR to automate this. I think @amureki did a great job with `django_dramatiq`'s workflow (see [here](https://github.com/Bogdanp/django_dramatiq/blob/master/.github/workflows/release.yml)) so maybe we could just port that over here.

jenstroeger:
> I'd happily accept a PR to automate this. I think @amureki did a great job with `django_dramatiq`'s workflow (see [here](https://github.com/Bogdanp/django_dramatiq/blob/master/.github/workflows/release.yml)) so maybe we could just port that over here.

[...]

PR #532, opened 2023-02-27
Add a release GHA workflow using 'build'

sirosen:
- release-published triggers a release to PyPI
- the build is done via pypa's `build` which provides a generic build frontend even if the project moves off of setuptools in the future
- upload is done by the pypa-provided github action

References:
- #531 motivated this
- credit to @amureki for the original workflow on which this is modeled


Bogdanp:
Thanks!

Issue #238, opened 2019-11-16
Async/await


peterdeme:
Hello guys. Just a question. Is there any reason why there's no async support for actors?

Bogdanp:
Unlikely to ever happen. I'm [not a fan](https://news.ycombinator.com/item?id=18110319) of asyncio as done in Python.

peterdeme:
I’m a bit confused why you are so against it because that can be implemented in a backward compatible way.

```python
if type(actor) is Coroutine:
    loop = asyncio.get_event_loop()
    loop.run_until_finished(actor())
else:
    result = actor()
```

this alone would be a huge improvement to your library for anyone who uses async ORMs or async HTTP libs.

rafalp:
Increasing number of codebases and python libraries adapt asyncio. Currently its possible to use Dramatiq and Celery with those by wrapping one's tasks with asgi_ref's `async_to_sync`, but it would be nice if this boilerplate lived in the task runner instead.

levrik:
I agree. We're using GraphQL subscriptions with Ariadne framework which requires `asyncio` and thus `asyncio`-ready libraries. Now we want to push additional events from our background jobs to a Redis channel so a GraphQL subscription can push this to the user in realtime. This needs the use of an async Redis client which doesn't work in Dramatiq if we don't wrap all our jobs in the workaround posted by @peterdeme above.
The other option is to use the plain sync `redis` client but I don't really want to pull in 2 different libraries into one application.

Bogdanp:
Anybody is welcome to try and contribute this. If it truly ends up not being complicated, it will get merged. On the worker side, I agree that it'll likely be about as simple as @peterdeme points out. On the enqueueing side, it doesn't seem possible except by cooperating with some sort of thread pool (either provided by the user or by bringing our own).

giuliotosato:
any news?

levrik:
@giuliotosato You can just define async actors since `1.15.0` (https://github.com/Bogdanp/dramatiq/releases/tag/v1.15.0) release in October 2023.

PR #536, opened 2023-03-06
Support asynchronous (asyncio) contexts in actors

caspervdw:
Hi @Bogdanp 

Thanks for the great work on this package!

I ported some code from another project to support asyncio with dramatiq. Reading #238 I got the feeling there might be people wanting to use this. We ourselves have an async webserver, of which we wanted to use parts from a worker process, and `asgiref.sync_to_async` didn't play nice with `sqlalchemy`, so we made this.

The bottomline here is to have a single event loop thread *per worker process*. This `EventLoopThread` is spinned up from the `AsyncMiddleware`. If `AsyncMiddleware` is inplace, you can decorate coroutines with `@async_actor`. Worker threads then schedule on the event loop and wait for their tasks to finish. In this way, I didn't have to touch the dramatiq core code.

There are some points to discuss here, but first I'd like to know if @Bogdanp is :+1: on proceeding with this PR?

### DIFF
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -18,7 +18,8 @@ from __future__ import annotations
 
 import re
 import time
-from typing import TYPE_CHECKING, Any, Callable, Dict, Generic, Optional, TypeVar, Union, overload
+from inspect import iscoroutinefunction
+from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, Generic, Optional, TypeVar, Union, overload
 
 from .broker import Broker, get_broker
 from .logging import get_logger
@@ -51,10 +52,9 @@ class Actor(Generic[P, R]):
       options(dict): Arbitrary options that are passed to the broker
         and middleware.
     """
-
     def __init__(
         self,
-        fn: Callable[P, R],
+        fn: Callable[P, Union[R, Awaitable[R]]],
         *,
         broker: Broker,
         actor_name: str,
@@ -63,7 +63,12 @@ class Actor(Generic[P, R]):
         options: Dict[str, Any],
     ) -> None:
         self.logger = get_logger(fn.__module__, actor_name)
-        self.fn = fn
+        if iscoroutinefunction(fn):
+            from dramatiq.middleware.asyncio import async_to_sync
+
+            self.fn = async_to_sync(fn)
+        else:
+            self.fn = fn  # type: ignore
         self.broker = broker
         self.actor_name = actor_name
         self.queue_name = queue_name
--- a/dramatiq/middleware/asyncio.py
+++ b/dramatiq/middleware/asyncio.py
@@ -1,0 +1,152 @@
+import asyncio
+import threading
+import time
+from concurrent.futures import TimeoutError
+from typing import Awaitable, Optional
+
+import dramatiq
+from dramatiq.middleware import Middleware
+
+from ..logging import get_logger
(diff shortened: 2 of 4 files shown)


Bogdanp:
I've only had a chance to skim this so far, but it looks great! I'll do a more in-depth review when I have more time in a couple of weeks. Thanks!

caspervdw:
OK great that you are positive on the approach. Two points I'd like to discuss:

- It would be leaner if we got rid of the `async_actor` and put an `inspect.iscoroutinefunction` inside the base `Actor` class. That would recognize `async def` functions, check if the asyncmiddleware is in place, and wrap the coroutine.
- I am not happy with the `broker.run_coroutine` monkeypatch. But I am not sure what to do then; we need something that is accessible from other middleware (say you want to initialize your database connection pool *on the event loop*), and from the actor. Maybe you have ideas how to fit this into the architecture?

Bogdanp:
> It would be leaner if we got rid of the async_actor and put an inspect.iscoroutinefunction inside the base Actor class. That would recognize async def functions, check if the asyncmiddleware is in place, and wrap the coroutine.

I think this would be fine.

> I am not happy with the broker.run_coroutine monkeypatch. But I am not sure what to do then; we need something that is accessible from other middleware (say you want to initialize your database connection pool on the event loop), and from the actor. Maybe you have ideas how to fit this into the architecture?

Would it work to just have a shared `event_loop_thread` in a module + `{get,set}_event_loop_thread` functions like we have with brokers? Why is the monkeypatching necessary?

Bogdanp:
Thanks! I've squashed your changes and made some of my own on top.

caspervdw:
Thanks @Bogdanp for finishing, the result looks great.

dream2333:
Is it possible to provide a way to change the default event loop? Using uvloop can further improve asynchronous efficiency, here are some of my benchmarks:

Python 3.9	Python 3.11
gevent	34281.80 requests / second	32258.07 requests / second
asyncio	40144.52 requests / second	51652.89 requests / second
asyncio + uvloop	64102.57 requests / second	66577.90 requests / second

Bogdanp:
@dream2333 yes, the event loop itself can probably be parameterized. Alternatively, it seems like it would be relatively easy for folks to extend `EventLoopThread` and the `AsyncIO` middleware in their own project.

MrFox131:
Hi! Are there any estimations when this will come to new release?

denjas:
Could you show a working example?
I have 2 files:
```python
# tasks.py
import requests
import dramatiq

@dramatiq.actor
async def count_words(url):
    response = requests.get(url)
    count = len(response.text.split(" "))
    print(f"There are {count} words at {url!r}.")
```
```python
#main.py
import asyncio
from tasksd import count_words

async def start():
    count_words.send("http://example.com")

if __name__ == '__main__':
    asyncio.run(start())
```
I launch the worker
```bash
dramatiq --processes 1 tasks
```
I'm adding a task
```bash
 python  main.py
 ```
 I get an error in the worker:
 ```bash
 RuntimeError: Global event loop thread not set. Have you added the AsyncIO middleware to your middleware stack?
 ```
Please tell me what am I doing wrong? Or provide a link to a working example?

ArtemIsmagilov:
Your code use sync pika. Does it make sense to use aio-pika https://github.com/mosquito/aio-pika ?

PR #544, opened 2023-04-07
Fix “an interrupted (Cancelled) coroutine doesn't cleanup”

caspervdw:
Following up #536 ; it seems that there is no cleanup happening when an asyncio actor is interrupted. It is often important to leave stuff in a consistent state (e.g. releasing a lock) when interrupting an actor.

This does not happen correctly as soon as there is an `await` in the `finally` block.

~I have no clue yet how to approach this issue.~

Redoing the `run_courotine_threadsafe` turned out to be quite cumbersome. So in the end I solved the issue by adding a "done" event and wrapping the coroutine in a `try .. finally`.

### DIFF
--- a/dramatiq/asyncio.py
+++ b/dramatiq/asyncio.py
@@ -171,10 +171,14 @@ class EventLoopThread(threading.Thread):
                     continue
                 else:
                     break
-        except Interrupt:
+        except Interrupt as e:
             # Asynchronously raised from another thread: cancel
-            # the future and reiterate to wait for possible
-            # cleanup actions.
+            # the future and wait for possible cleanup actions.
             self.loop.call_soon_threadsafe(task.cancel)
+            try:
+                future.result()
+            except asyncio.CancelledError:
+                pass
+            raise e  # raises Interrupt, not asyncio.CancelledError
 
         return future.result()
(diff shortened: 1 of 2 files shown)


caspervdw:
@Bogdanp  Please review! Thanks in advance.

Bogdanp:
I was planning to review this weekend, but it looks like I won't have time.  I'll try to do it in two weeks.

caspervdw:
@Bogdanp Do you think you can find time to review this?

Bogdanp:
Sorry @caspervdw, I've got a reminder set up to look at this, but no time so far. Probably the earliest I'll be able to do it is sometime around the end of next week.

dramatiq/asyncio.py
```diff
@@ -129,16 +129,28 @@ def run_coroutine(self, coro: Awaitable[R]) -> R:
...
+                    return future.result(timeout=self.interrupt_check_ival)
+                except concurrent.futures.TimeoutError:
+                    continue
+        except Interrupt:
+            # Asynchronously raised from another thread: cancel the future.
+            self.loop.call_soon_threadsafe(future.cancel)
+            # Wait for the 'done' event instead of the future; the future will
+            # raise CancelledError immediately while we should wait for the coro
+            # to actually finish its cleanup actions.
+            done.wait()
```
Bogdanp, line 155: This should probably have a timeout, and raise another error when it does time out. Otherwise, the changes look good.

PR #545, opened 2023-04-10
Mark message as failed when fail to process

jairhenrique:
Mark message as `failed` when actor raises an Exception.

On `self.consumers[message.queue_name].post_process_message(message)` the method `nack` will be called.

This change will enable refactor on `dramatiq_sqs` to skip message delete and redrive message to dlq. https://github.com/Bogdanp/dramatiq_sqs/issues/10#issuecomment-1502280022

### DIFF
--- a/dramatiq/worker.py
+++ b/dramatiq/worker.py
@@ -503,6 +503,7 @@ class _WorkerThread(Thread):
 
         except BaseException as e:
             message.stuff_exception(e)
+            message.fail()
 
             throws = message.options.get("throws") or (actor and actor.options.get("throws"))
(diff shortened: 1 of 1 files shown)


spumer:
Without that fix we have incorrect behaviour in Retries middleware, when enqueue to DELAY queue failed we silently ACK message and lose it

@Bogdanp

spumer:
It's would not be redelivered since it's acked. 

- processing failed
- retry middleware try to delay
- delay failed
- message acked as processed

As result we have error but no have message in dead letter queue or delay queue.

PR #546, opened 2023-04-14
Ensure only canonical are added to RabbitmqBroker.queues

es:
When a worker is started, it'll eventually start two consumers: one for reading off the queue and another for reading off the delayed queue. When it starts consuming off the delayed queue, it calls `RabbitmqBroker#consume('queue_name.DQ')`. In turn, `queue_name.DQ` goes through the declare queue initialization which results in its name being added to the `RabbitmqBroker.queues` set. This becomes an issue as later on in commands such as `RabbitmqBroker#flush`, where assumptions are made that only the canonical queue name is present.

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -26,7 +26,7 @@ from threading import Event, local
 import pika
 
 from ..broker import Broker, Consumer, MessageProxy
-from ..common import current_millis, dq_name, xq_name
+from ..common import current_millis, dq_name, q_name, xq_name
 from ..errors import ConnectionClosed, DecodeError, QueueJoinTimeout
 from ..logging import get_logger
 from ..message import Message, get_encoder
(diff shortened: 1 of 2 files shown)


es:
This is part of a series of PRs leading up to the addition of Quorum Queues. Next in the series are:
- https://github.com/wavemm/dramatiq/pull/1
- https://github.com/wavemm/dramatiq/pull/2

@Bogdanp let me know how you'd like to approach the set of PRs and if there are any larger issues you'd like me to address, thank you!

Bogdanp:
Thanks!

Regarding the other PRs, they also seem fine, though they both needs docs and the change to add an argument to drop queues to `flush` I think should be replaced with a new method to drop a queue. It's fine if the method is RMQ-specific.

aW3st:
@es @Bogdanp We use dramatiq at my company and would love to have quorum queues available, in particular for poison message handling. Anything we can do to move this story forward?

PR #549, opened 2023-04-22
build: add missing asyncio module to package

scabala:
I've encountered issue when installing Dramatiq from commit `78f0f8bbd6312ebb0b811f661ef5fe8773107aa3`. 

It turns out that asyncio support was added but module that implements this supports is not included in distribution package. It fixes error `ModuleNotFoundError: No module named 'dramatiq.asyncio'`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,6 +106,7 @@ setup(
     long_description_content_type="text/markdown",
     packages=[
         "dramatiq",
+        "dramatiq.asyncio",
         "dramatiq.brokers",
         "dramatiq.middleware",
         "dramatiq.rate_limits",


Bogdanp:
Thank you!

Bogdanp:
I should've waited to drink my coffee before merging.

Can you post the full stack trace here of the issue you're running into? `dramatiq.asyncio` is not a package, so I'm going to have to revert this change. The module should be included in the distribution so it's weird that you're getting an error.

scabala:
Hi @Bogdanp 
I'm sorry about the situation - as I've mentioned, I've encountered `ModuleNotFound` error when trying to import `dramatiq.asyncio`. However, it must have been something with my local setup since `asyncio.py` is indeed part of installed package.

Once again, sorry for the trouble.

Bogdanp:
Oh, don't worry about it! I didn't mean to imply you did something wrong. Mistakes happen and I was just trying to be funny, because I should've noticed the problem before merging.

It's still not clear to me, though, if the issue is gone for you or if there is still some problem. So, let me know if there is, and please provide a full stack trace if so.

PR #552, opened 2023-06-01
Don't reload on watchdog 'opened' events

seanpile:
As of watchdog >= 2.3, file open events on certain platforms are causing an 'opened' event to be delivered.  We should not be restarting the server on these changes, as they can be spammy and disruptive.

I was able to confirm the behaviour by running the tests within a debian-like environment (python slim bullseye); without the fix, the `tests/test_watch.py` would fail for the first test case (the evented observer)

For more context, see discussions here:

- https://github.com/gorakhargosh/watchdog/issues/949
- https://github.com/pallets/werkzeug/pull/2604/files

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -13,8 +13,16 @@ Added
 
 * Asyncio support.  (`#536`_, `@caspervdw`_)
 
+Changed
+^^^^^^^
+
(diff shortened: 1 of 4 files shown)


Bogdanp:
Thanks!

PR #556, opened 2023-06-29
fix: detect the missing of rabbitmq queue and redeclare it.

JetDrag:
The declared rabbitmq queue will be missing sometime, such as deletion by mistake or rabbitmq server failure. So I add some logic to make sure the missing queue to be declared again, as the recent implementation won't auto re-declare the missing queue.

There are two situation to re-declare the missing queue:

- Declare the queue every time when the consumer start.
- Redeclare the queue when queue is found to be missing during enqueue.

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -24,7 +24,6 @@ from itertools import chain
 from threading import Event, local
 
 import pika
-from pika.exceptions import ChannelClosedByBroker
 
 from ..broker import Broker, Consumer, MessageProxy
 from ..common import current_millis, dq_name, q_name, xq_name
@@ -309,7 +308,8 @@ class RabbitmqBroker(Broker):
           ConnectionClosed: If the underlying channel or connection
             has been closed.
         """
-        queue_name = message.queue_name
+        canonical_queue_name = message.queue_name
+        queue_name = canonical_queue_name
 
         if delay is not None:
             queue_name = dq_name(queue_name)
@@ -324,7 +324,7 @@ class RabbitmqBroker(Broker):
         attempts = 1
         while True:
             try:
-                self.declare_queue(queue_name, ensure=True)
+                self.declare_queue(canonical_queue_name, ensure=True)
                 self.logger.debug("Enqueueing message %r on queue %r.", message.message_id, queue_name)
                 self.emit_before("enqueue", message, delay)
                 self.channel.basic_publish(
@@ -344,13 +344,6 @@ class RabbitmqBroker(Broker):
                 # Delete the channel and the connection so that the
                 # next caller/attempt may initiate new ones of each.
                 del self.connection
-
-                # When a queue has been deleted, attempt to get it
-                # redeclared by removing it from the known set. The next
-                # time a message is enqueued -- which could be when we
-                # retry this block -- it will be redeclared.
-                if isinstance(e, ChannelClosedByBroker) and e.reply_code == 404:
-                    self.queues.remove(q_name(queue_name))
 
                 attempts += 1
                 if attempts > MAX_ENQUEUE_ATTEMPTS:
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -86,8 +86,7 @@ class RabbitmqBroker(Broker):
     .. _ConnectionParameters: https://pika.readthedocs.io/en/stable/modules/parameters.html
     """
 
-    def __init__(self, *, confirm_delivery=False, url=None, middleware=None, max_priority=None, parameters=None,
-                 **kwargs):
+    def __init__(self, *, confirm_delivery=False, url=None, middleware=None, max_priority=None, parameters=None, **kwargs):
         super().__init__(middleware=middleware)
 
         if max_priority is not None and not (0 < max_priority <= 255):
@@ -218,7 +217,7 @@ class RabbitmqBroker(Broker):
         Returns:
           Consumer: A consumer that retrieves messages from RabbitMQ.
         """
-        self.declare_queue(queue_name, ensure="strict")
+        self.declare_queue(queue_name, ensure=True)
         return self.consumer_class(self.parameters, queue_name, prefetch, timeout)
 
     def declare_queue(self, queue_name, *, ensure=False):
@@ -227,9 +226,8 @@ class RabbitmqBroker(Broker):
 
         Parameters:
           queue_name(str): The name of the new queue.
-          ensure(bool|str): When True, the queue is created immediately on
-            the server. When 'strict', make sure the queue is created on the
-            server.
+          ensure(bool): When True, the queue is created on the server,
+            if necessary.
 
         Raises:
           ConnectionClosed: When ensure=True if the underlying channel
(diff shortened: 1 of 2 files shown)


JetDrag:
Hi, @Bogdanp , do you have time to review this fix? Many thanks to your work!

JetDrag:
Hi, @Bogdanp , do you have time to review this fix?  I've update a newer version.  `:grin:`

Bogdanp:
Hi @JetDrag, I'll try to review this PR sometime around the end of next week.

JetDrag:
@Bogdanp roger, thanks

dramatiq/brokers/rabbitmq.py
```diff
@@ -216,17 +218,19 @@ def consume(self, queue_name, prefetch=1, timeout=5000):
         Returns:
           Consumer: A consumer that retrieves messages from RabbitMQ.
         """
-        self.declare_queue(queue_name, ensure=True)
+        self.declare_queue(queue_name, ensure=True, strict=True)
         return self.consumer_class(self.parameters, queue_name, prefetch, timeout)
 
-    def declare_queue(self, queue_name, *, ensure=False):
+    def declare_queue(self, queue_name, *, ensure=False, strict=False):
```
Bogdanp, line 224: Overall, this looks like a fine change, but I'm wondering if, instead of having `ensure` and `strict` be separate flags, we shouldn't just extend `ensure` to `bool | "strict"`.

dramatiq/brokers/rabbitmq.py
```diff
@@ -344,6 +348,11 @@ def enqueue(self, message, *, delay=None):
                 # next caller/attempt may initiate new ones of each.
                 del self.connection
 
+                if isinstance(e, ChannelClosedByBroker) and e.reply_code == 404 and \
+                        e.reply_text.startswith("NOT_FOUND - no queue"):
```
Bogdanp, line 352: Is the string matching necessary here or can we rely on the code alone?

JetDrag:
Hi, @Bogdanp , I've resolve the conversation you added before, please review again.

tests/test_rabbitmq.py
```diff
@@ -464,6 +465,38 @@ def do_work():
...
+    # And I send that actor a message
+    do_work.send()
+    try:
+        rabbitmq_broker.join(do_work.queue_name, timeout=20000)
+        worker.join()
+
+        # Then the queue should eventually be declared and the message executed
+        assert executed
+    finally:
+        worker.stop()
```
spumer, line 497: Looks like try/finally is redundant

```suggestion
    # And I send that actor a message
    do_work.send()
    rabbitmq_broker.join(do_work.queue_name, timeout=20000)
    worker.join()
    worker.stop()

    # Then the queue should eventually be declared and the message executed
    assert executed
```

JetDrag: Be ensure to cover the exception then close the worker safely during the test. I reserve the try/finally scope. But I move the assert statement to the end of the test case.

JetDrag: Be ensure to cover the exception then close the worker safely during the test. I reserve the try/finally scope. But I move the assert statement to the end of the test case.

JetDrag:
Hi, @Bogdanp , I've resolve the conversation you added before,  when will you have time to review the renewed fix?

We have run Dramatiq with this fix for a while at our production environment and it's necessary for long-running rabbitmq workers to include this fix to ensure reliability.

Many thanks!

JetDrag:
Hi, @Bogdanp , I've resolve the failure of lint, make sure the test cases in workflow is passed in my forked project. Could you approve the workflow in this PR?

Bogdanp:
I see both my comments were marked as resolved, but they don't seem to be resolved?

dramatiq/brokers/rabbitmq.py
JetDrag: corrected

dramatiq/brokers/rabbitmq.py
JetDrag: the code is enough

JetDrag:
> I see both my comments were marked as resolved, but they don't seem to be addressed?

Sorry, there's something missing during rebase. I've updated the code to make sure your comments are resloved.

Bogdanp:
The suggestion in https://github.com/Bogdanp/dramatiq/pull/556/files#r1334929506 still doesn't seem to be addressed. The idea is to avoid adding a `strict` argument,  and instead make `ensure` take `True`, `False` or `"strict"`.

JetDrag:
Sorry for the omission, it seem there're some bugs in the online github CodeSpace. I've refined and pushed it again.

Thank you for your patience and work.

dramatiq/brokers/rabbitmq.py
```diff
@@ -243,13 +246,13 @@ def declare_queue(self, queue_name, *, ensure=False):
...
-            self._ensure_queue(queue_name)
+            self._ensure_queue(queue_name, ensure == "strict")
 
-    def _ensure_queue(self, queue_name):
+    def _ensure_queue(self, queue_name, strict=False):
         attempts = 1
         while True:
             try:
-                if queue_name in self.queues_pending:
+                if strict or queue_name in self.queues_pending:
```
Bogdanp, line 255: Taking another look at this, I'm not sure the strict argument is even required. Since the block at line 350 adds the missing queue to the `queues_pending` set, when the next message is enqueued, the queue will be created anyway, won't it? 

Separately, defaulting `ensure` to `"strict"` will cause the queue to be redeclared on every enqueue, which seems bad. If possible, I think we should drop the `strict` stuff altogether, or default `ensure` to `True` as before.

JetDrag: The `ensure` value during enqueue is still `True`, the queue won't be declared actually every time util it's added to `queues_pending` set by some error handle logic.

I also move `self.declare_queue(queue_name, ensure=True)` to enqueue retry loop, make sure user won't retry again by themselves to redeclared the queue.

The `strict` is set only during consumer starting to ensure the queue is redeclared. 
Because there's no extra broker specific handle logic which can be implemented in consumer to handle with the redeclared of rabbitmq queue, and the consumer won't start/restart many time.

JetDrag: Hello, @Bogdanp, could you take some time to review this response? Thanks to your patient and meticulous!

dramatiq/brokers/rabbitmq.py
```diff
@@ -324,6 +324,7 @@ def enqueue(self, message, *, delay=None):
         attempts = 1
         while True:
             try:
+                self.declare_queue(queue_name, ensure=True)
```
Bogdanp, line 327: This change is wrong when `delay` is not None.

tests/test_rabbitmq.py
```diff
@@ -464,6 +466,50 @@ def do_work():
             worker.stop()
 
 
+def test_rabbitmq_broker_retries_declaring_queues_when_declared_queue_disappears(rabbitmq_broker):
```
Bogdanp, line 469: This test doesn't actually cover any of the new code

JetDrag:
Awesome work! Thanks to your patient and meticulous!

PR #561, opened 2023-07-19
chore: add type hint to global dramatiq.broker.global_broker variable

### LINKED ISSUE #555: Type issue: statements become unreachable when using `global_broker`
# Issues

With the latest [version 1.4 of mypy](https://mypy-lang.blogspot.com/2023/06/mypy-140-released.html) some of my type checks are failing. For example:
```python
broker = dramatiq.broker.global_broker
assert broker is not None  # This is to satisfy mypy.

broker.enqueue(message, delay=delay)  # type: ignore[no-untyped-call]
 ```
now gives me for the `enqueue()` call
```
error: Statement is unreachable  [unreachable]
```
I think that’s because the `global_broker` is considered `None`

https://github.com/Bogdanp/dramatiq/blob/43b498f0afde90696a433631f947ec82780ca987/dramatiq/broker.py#L24-L25

which I think is also incorrect by mypy. However, this change:
```python
#: The global broker instance.
global_broker: Broker | None = None
```
fixes the issue; or use [Optional[]](https://docs.python.org/3/library/typing.html#typing.Optional).

## What OS are you using?

macOS Ventura 13.4

```
>>> dramatiq.__version__
'1.14.2'
```

## What did you do?

Me? Nothing! 🤓

Type checks should continue to pass with new mypy.

## What happened?

As above, mypy fails.

jenstroeger:
Fixes #555

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -58,3 +58,4 @@ of those changes to CLEARTYPE SRL.
 | [@cmflynn](https://github.com/cmflynn)                 | Cody Flynn             |
 | [@dancardin](https://github.com/dancardin)             | Dan Cardin             |
 | [@caspervdw](https://github.com/caspervdw)             | Casper van der Wel     |
+| [@jenstroeger](https://github.com/jenstroeger/)        | Jens Troeger           |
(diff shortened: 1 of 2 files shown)


jenstroeger:
@Bogdanp looks like CI workflows are broken, can you please take a look?

PR #562, opened 2023-08-08
Improved legibility of the help parameter for --processes

santigandolfo:
Currently when we run `dramatiq --help`,  the help information for --processes says:

`the number of worker processes to run (default: %s)`, where %s varies for the number of CPUs in the system that run the command.

The problem is that it is an ambiguous hint, because if one does't know how the value was internally defined, they could wrongly asume that the number that appears as a default is a hardcoded one.

This PR intends to make it clear that the value is not hardcoded, so if a user wanted a specific number of --processes in all the systems where their code is run, then they should set the value for it


santigandolfo:
> Currently when we run `dramatiq --help`, the help information for --processes says:
> 
> `the number of worker processes to run (default: %s)`, where %s varies for the number of CPUs in the system that run the command.
> 
> The problem is that it is an ambiguous hint, because if one does't know how the value was internally defined, they could wrongly asume that the number that appears as a default is a hardcoded one.
> 
> This PR intends to make it clear that the value is not hardcoded, so if a user wanted a specific number of --processes in all the systems where their code is run, then they should set the value for it

@Bogdanp could you check this oneliner PR? It's a really small improvement that could help other people

Bogdanp:
Thanks, but I think the current message is fine as-is.

PR #563, opened 2023-08-15
Use backend from group/pipeline broker in get_result(s)

DiegoPomares:
These are the proposed changes to address the issue described in: https://github.com/Bogdanp/dramatiq/issues/564

The issue is that the groups and pipelines look for the backend of the default/global broker when getting results, regardless of the broker they were declared with.


Bogdanp:
Thanks!

Issue #522, opened 2023-01-05
FeatReq: Support timedelta for delays?


dragonpaw:
In GoLang they have a Duration class that is actually pretty nifty. Instead of passing a int of seconds to sleep, you pass a Duration. Python's similar timedelta class doesn't get much love, but I enjoy using it in place of ints for time within my own code a lot. I notice in Dramatiq, the code wants timeouts and delays passed as millis, which makes sense, but isn't intuitive perhaps from the hinting. So I was wondering if Dramatiq could also accept timedeltas and use those, as there's no ambiguity in them. (While still treating ints as millis for backwards compatibility.)

It'd look something like this:
```python
actor.send_with_options(args=(url), delay=timedelta(seconds=2))
```

And within the code it's just:
```python
def send_with_options(
    self, *,
    args: tuple = (),
    kwargs: Optional[Dict[str, Any]] = None,
    delay: Optional[Union[datetime.timedelta,int]] = None,
    **options,
) -> Message[R]:
    if isinstance(delay, timedelta):
        delay = delay.total_seconds() * 1000
```

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/Bogdanp/dramatiq/issues/522">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/Bogdanp/dramatiq/issues/522/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/Bogdanp/dramatiq/issues/522/pledge.svg">
</picture>
</a>

odigity:
I like it.

Since you've already written the code, how about opening a PR for it?  (and maybe a test or two)  Looks like the maintainer already has his hands full...

Bogdanp:
I think this is a good idea and a PR would be appreciated.

arseniiarsenii:
Hi, @h3nnn4n, I noticed, that the changes are not consistent with other parts of the project which also accept delays in milliseconds. timedelta should also be acceptable as:

- actor's min_backoff parameter
- actor's max_backoff parameter
- dramatiq.errors.Retry exception's 'delay' argument

These are some places that come to my mind, there are probably more of them.

Otherwise, thanks for the PR!

Bogdanp:
[...]

PR #569, opened 2023-08-20
Add support for timedelta as actor delay

h3nnn4n:
This implements https://github.com/Bogdanp/dramatiq/issues/522 as suggested by @dragonpaw

Support for using `timedelta` for the actor delay was added at the public interface. Alternatively it could have been implemented by switching to timedelta internally and converting to int / float when interfacing with external libraries. This would improve type safety in dramatiq, but it is a more intrusive change.

Based on feedback from this PR I can open a followup to support `timedelta` for timeout args as well.


Bogdanp:
Thanks!

PR #582, opened 2023-10-10
feature(redis): Dequeue a single message

theilgaard:
Addresses https://github.com/Bogdanp/django_dramatiq/issues/24

### DIFF
--- a/dramatiq/brokers/redis/dispatch.lua
+++ b/dramatiq/brokers/redis/dispatch.lua
@@ -162,6 +162,17 @@ if command == "enqueue" then
     redis.call("rpush", queue_full_name, message_id)
 
 
+-- Dequeues a message on $queue_full_name.
+if command == "dequeue" then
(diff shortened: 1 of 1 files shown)


dramatiq/brokers/redis/dispatch.lua
```diff
@@ -162,6 +162,17 @@ if command == "enqueue" then
     redis.call("rpush", queue_full_name, message_id)
 
 
+-- Dequeues a message on $queue_full_name.
+if command == "dequeue" then
+    local message_id = ARGS[1]
+    local n = redis.call("lrem", queue_full_name, message_id)
```
Qu4tro, line 168: I think it's missing the count arg [1].

My suggestion would be to change it to:
`local n = redis.call("lrem", queue_full_name, 0, message_id)`

[1] https://redis.io/commands/lrem/

PR #583, opened 2023-10-25
Replaced linting tools with with ruff and added ruff formatter

santigandolfo:
Currently, the repo uses isort, flake8, flake8-bugbear and flake8-quotes. All of these tools can be replaced by ruff, which not only has even more tools included but also executes linting and formatting much faster.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.1.2
+    hooks:
+      - id: ruff
+        args: [ --fix, --unsafe-fixes, --exit-non-zero-on-fix ]
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.1.2
+    hooks:
+      - id: ruff-format
```
bonastreyair, line 20: updated to newer version + single pre-commit repo
```suggestion
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.1.8
    hooks:
      - id: ruff-format
      - id: ruff
        args: [--fix, --exit-non-zero-on-fix]
```

dramatiq/actor.py
```diff
@@ -54,15 +54,16 @@ class Actor(Generic[P, R]):
       options(dict): Arbitrary options that are passed to the broker
         and middleware.
     """
+
     def __init__(
         self,
-        fn: Callable[P, Union[R, Awaitable[R]]],
+        fn: Callable[P, R | Awaitable[R]],
```
bonastreyair, line 60: be aware this things will make older python version to be incompatible too...

santigandolfo: @bonastreyair what do you think?

Bogdanp:
Thanks, but this is too big of a change, and the current tools are fine.

PR #584, opened 2023-10-29
redis: add support for broker priority

davidt99:
Create different queues for priority steps and consuming them in that order.
I took a similar approach as celery's implementation of priority queues in redis.


davidt99:
I'll close this for now since the build is failing, will reopen when it's ready

Issue #558, opened 2023-07-07
OSError: [Errno 98] Address already in use when calling send function


heatherKoo07:
# Issues

GitHub issues are for bugs.  If you have questions, please ask them on the [mailing list](https://groups.io/g/dramatiq-users/topics).

## Checklist

* [ ] Does your title concisely summarize the problem?
* [ ] Did you include a minimal, reproducible example?
* [ ] What OS are you using?
* [ ] What version of Dramatiq are you using?
* [ ] What did you do?
* [ ] What did you expect would happen?
* [ ] What happened?

## What OS are you using?

Debian 12

## What version of Dramatiq are you using?

v1.12.0

## What did you do?

I'm using dramatiq in a flask app with flask-melodramatiq. I got an OSError from POST request where dramatiq workers are supposed to run an async job. 

## What did you expect would happen?

The task is supposed to be processed successfully

## What happened?
The below error was raised and tasks in queue were not processed but remained for hours. I had to restart flask to solve this issue.

```
Jun 24 02:09:32 c14-irkif docker-kif-web: msg = actor.send(vm, **user_arg_map)
Jun 24 02:09:32 c14-irkif docker-kif-web: File "/usr/local/lib/python3.8/site-packages/dramatiq/actor.py", line 112, in send
Jun 24 02:09:32 c14-irkif docker-kif-web: return self.send_with_options(args=args, kwargs=kwargs)
Jun 24 02:09:32 c14-irkif docker-kif-web: File "/usr/local/lib/python3.8/site-packages/dramatiq/actor.py", line 131, in send_with_options
Jun 24 02:09:32 c14-irkif docker-kif-web: return self.broker.enqueue(message, delay=delay)
Jun 24 02:09:32 c14-irkif docker-kif-web: File "/usr/local/lib/python3.8/site-packages/dramatiq/brokers/rabbitmq.py", line 319, in enqueue
Jun 24 02:09:32 c14-irkif docker-kif-web: self.channel.basic_publish(
Jun 24 02:09:32 c14-irkif docker-kif-web: File "/usr/local/lib/python3.8/site-packages/dramatiq/brokers/rabbitmq.py", line 153, in channel
Jun 24 02:09:32 c14-irkif docker-kif-web: channel = self.state.channel = self.connection.channel()
Jun 24 02:09:32 c14-irkif docker-kif-web: File "/usr/local/lib/python3.8/site-packages/dramatiq/brokers/rabbitmq.py", line 125, in connection
[...]

PR #589, opened 2023-11-06
Fix RabbitMQ connection leak

heatherKoo07:
This is a fix for the issue https://github.com/Bogdanp/dramatiq/issues/558

Problem is RabbitMQ connection is leaking in dramatiq. When messages are enqueued to RabbitMQ, the `pika.BlockingConnection`s are not closed after the messages are delivered. This eats up all the local socket ports under high load and raises `Address already in use` error.

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -352,6 +352,8 @@ class RabbitmqBroker(Broker):
                     "Retrying enqueue due to closed connection. [%d/%d]",
                     attempts, MAX_ENQUEUE_ATTEMPTS,
                 )
+            finally:
+                del self.connection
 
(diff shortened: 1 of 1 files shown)


Bogdanp:
I am reluctant to merge this since, for folks who don't use gevent (i.e. most people), this wouldn't be a problem as the broker would only create one connection per OS thread instead of one connection per green thread.

Does a middleware like

```python
class CloseMiddleware(Middleware):
    def after_enqueue(self, broker, message, delay):
        del broker.connection
```

not resolve the issue for you? It should effectively be doing the same thing as this PR, except without modifying Dramatiq.

heatherKoo07:
Thank you for the suggestion! The CloseMiddleware seems to resolve the issue for me.

PR #592, opened 2023-12-09
Remove results message key dependency from actor and queue

SYNchroACK:
The idea here is to allow the possibility to get a result only based on a `message_id` and `namespace` like the following example:

**app.py**
```
results = RedisBackend(host="localhost", namespace="my-results")
```

**endpoint.py**: fastapi endpoint to collect result of a long task
```
from .app import results

@router.get("/{message_id}")
async def get(message_id: str):
    message = Message(
        queue_name="",
        actor_name="",
        message_id=message_id,
        args=(),
        kwargs={},
        options={},
    )

    data = json.loads(results.get_result(message))
    
    return data
```

I believe removing `actor` and `queue` from the generation of message key will not affect in any way the system.

### DIFF
--- a/dramatiq/results/backend.py
+++ b/dramatiq/results/backend.py
@@ -143,10 +143,8 @@ class ResultBackend:
         Returns:
           str
         """
-        message_key = "%(namespace)s:%(queue_name)s:%(actor_name)s:%(message_id)s" % {
+        message_key = "%(namespace)s:%(message_id)s" % {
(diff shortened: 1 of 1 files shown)


Bogdanp:
Unfortunately, this would be a breaking change for existing users. I'd recommend subclassing your result backend and overriding the `build_message_key` implementation for your own use case.

PR #596, opened 2023-12-19
Improve handling of newlines in log messages

5tefan:
(Apologies for not starting a discussion as requested in contributing guidelines... let's have the discussion here and reject this PR if it doesn't align with your vision.)

## Summary

Improves the newline handling in log forwarding from workers so that it does not always introduce newlines.

### Problem

Due to some unknown change in our setup, we started seeing log messages from workers split into multiple lines:

image

Perhaps we are flushing log message more frequently than expected by dramatiq... and the flushes are not aligned with newlines. 

### Solution

After some digging, we found that the newlines are inserted by dramatiq here:

https://github.com/Bogdanp/dramatiq/blob/a30b7878c5f7570864b1e056580faecbf2abdad2/dramatiq/cli.py#L322-L338

 This code _always_ adds a newline, even if the `rstrip` did not strip one out.

Given the comment above this code section explaining the goal:

> discard newline-only data from the pipe

I believe this revision accomplishes this better by only using `rstrip` for the continue condition (to not pass blank lines),but otherwise passing the log without modification.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -63,3 +63,4 @@ of those changes to CLEARTYPE SRL.
 | [@DiegoPomares](https://github.com/DiegoPomares/)      | Diego Pomares          |
 | [@pahrohfit](https://github.com/pahrohfit/)            | Robert Dailey          |
 | [@nhairs](https://github.com/nhairs)                   | Nicholas Hairs         |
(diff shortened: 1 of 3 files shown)


Bogdanp:
Thanks! I'll try to review this PR this weeked or the next. The code in question does look like it assumes line buffering, so your setup probably changed to some other buffer mode for whatever reason.

PR #598, opened 2024-01-11
Watch patterns

### LINKED ISSUE #594: CLI: Allow watching custom include / exclude patterns  [feature-request]
First of all, great project :)

Currently the CLI reloader only allows specifying the base path path to watch for changes (`--watch path/to/watch`). However the watcher is configured so that it only listens to changes on `*.py` files ([source code](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/watcher.py#L16)).

I would like to be able to use the reloader with changes to other files such as configuration files.

It might look like the following:
```python
        # cli.py
        parser.add_argument(
            "-i",
            "--watch-include",
            action="append",
            dest="include_patterns",
            help="Additional patterns to watch for changes",
        )

        parser.add_argument(
            "-x",
            "--watch-exclude",
            action="append",
            dest="exclude_patterns",
            help="Patterns to ignore for changes",
        )
```

Which would then be used like:
```bash
dramatiq [...] --watch path/to/watch -i *.yml -i templates/*.html -x templates/*.png
```

Aside: in documentation / help strings it might be worth noting that prior to Python 3.13 these patterns only support multiple name (`**`) globbing at the start of the pattern ([ref](https://github.com/python/cpython/issues/73435#issuecomment-1569046011))

nhairs:
Fixes #594

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,4 +1,4 @@
-[#](#) Contributors
+# Contributors
 
 This file lists the contributors to the Dramatiq project.
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,4 +1,4 @@
-# Contributors
+[#](#) Contributors
 
 This file lists the contributors to the Dramatiq project.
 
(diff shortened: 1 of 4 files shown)


Bogdanp:
Thanks!

nhairs:
@Bogdanp - just wondering when you're likely to publish a new version of the package?

(So that I can make use of this change)

Bogdanp:
I’ve released 1.16 this morning.

PR #599, opened 2024-01-12
Added a `max_retries` option to Retry exception as it done with a `delay` option

dvazar:


Bogdanp:
Thanks, but I think this being configurable at the actor or message level is already enough.

PR #587, opened 2023-11-06
Add build support for python3.12, drop python3.7

FinnLidbetter:
Python3.12 has been released and python3.7 has been past end-of-life since 2023-06-27.
This change adds build support for python3.12 and drops build support for python3.7.

PR #606, opened 2024-01-25
Update python_requires in setup.py

nicoddemus:
Seems this bit was missed in https://github.com/Bogdanp/dramatiq/pull/587.


Bogdanp:
Thanks!

PR #619, opened 2024-04-11
feature: add middleware after_{consumer,worker}_thread_boot hooks

kamalmarhubi:
This allows the middleware to run code in the context of a worker or consumer thread before it enters its run loop. This could be used to set up thread-local resources, or as in the author's case, to get a reference to the thread before it does any work.

This was proposed on the mailing list [0] with Bogdan accepting the idea in principle [1].

[0]: https://groups.io/g/dramatiq-users/topic/105311701
[1]: https://groups.io/g/dramatiq-users/topic/105311701#258

### DIFF
--- a/dramatiq/middleware/middleware.py
+++ b/dramatiq/middleware/middleware.py
@@ -138,18 +138,22 @@ class Middleware:
         """Called after the worker process shuts down.
         """
 
+    def after_consumer_thread_boot(self, broker, thread):
+        """Called from a consumer thread after it starts but before it starts its run loop.
+        """
+
     def before_consumer_thread_shutdown(self, broker, thread):
         """Called before a consumer thread shuts down.  This may be
         used to clean up thread-local resources (such as Django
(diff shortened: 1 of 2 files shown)


kamalmarhubi:
converted to draft; will mark it ready for review after we've run our code using this middleware for a few days to be sure these hooks are adequate

kamalmarhubi:
@Bogdanp we've been runing a cherry-pick of this onto the version we use in prod and it's allowed us to start workers in a paused state as described in the mailing list.

Bogdanp:
Thanks!

kamalmarhubi:
@Bogdanp quick question: do you expect to make a release containing this any time soon? like I said we're running with a cherry-pick which is fine, mostly just want to have a mental note to do the upgrade if / when it's released. thanks!

Bogdanp:
@kamalmarhubi I plan to make a release sometime next week.

PR #629, opened 2024-06-12
Updated Retries middleware

kuba-lilz:
For motivation see discussion at 
https://groups.io/g/dramatiq-users/topic/how_to_check_how_long_message/106519634

Retries middleware now records requeue_timestamp on message options in `before_enqueue` function if message failed execution.

Hoped the implementation and tests are up to required standards, if not, let me know what needs to be changed.

### DIFF
--- a/dramatiq/middleware/retries.py
+++ b/dramatiq/middleware/retries.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import time
 import traceback
 
 from ..common import compute_backoff
@@ -114,3 +115,9 @@ class Retries(Middleware):
 
         self.logger.info("Retrying message %r in %d milliseconds.", message.message_id, delay)
         broker.enqueue(message, delay=delay)
+
+    def before_enqueue(self, broker, message, delay):
+
+        # If message will be retried, record in its options time at which it is requeued
+        if "retries" in message.options:
+            message.options["requeue_timestamp"] = int(time.time() * 1000)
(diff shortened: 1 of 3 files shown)


dramatiq/middleware/retries.py
```diff
@@ -114,3 +115,9 @@ def after_process_message(self, broker, message, *, result=None, exception=None)
 
         self.logger.info("Retrying message %r in %d milliseconds.", message.message_id, delay)
         broker.enqueue(message, delay=delay)
+
+    def before_enqueue(self, broker, message, delay):
+
+        # If message will be retried, record in its options time at which it is requeued
+        if "retries" in message.options:
+            message.options["requeue_timestamp"] = int(time.time() * 1000)
```
Bogdanp, line 123: Since you’re changing the Retries middleware directly, I would move this line under line 98 and avoid the check and separate before enqueue hook altogether.

kuba-lilz: My small concern was that maybe, just maybe, in corner cases there might be some lag between execution of that part of code and enqueue operation, given how dramatiq might be running workers doing heavy CPU operations at the same time.
I shouldn't worry about that?

Bogdanp: I think the most likely place for this thread to relinquish control is going to be the io that is going to be done when actually enqeueuing the message so I think the chances of the suggested approach being less reliable for timing are very slim (it might in fact be more reliable because it’s doing less work).

kuba-lilz: Roger that, I'll refactor and push again 🫡

Bogdanp:
Thanks!

PR #634, opened 2024-06-25
Close an asyncio event loop in the test suite

kurtmckee:
Noticed this warning while attempting to run the test suite with warnings enabled.

```
tests/middleware/test_asyncio.py::test_event_loop_thread_run_coroutine_interrupted
ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
```

This PR makes calls to `.stop()` join the thread and then close the event loop, which resolves the warning above.


Bogdanp:
Thanks!

Issue #640, opened 2024-07-12
Recognize actor with module name instead of only func name


z0z0r4:
# Issues

GitHub issues are for bugs.  If you have questions, please ask them on the [mailing list](https://groups.io/g/dramatiq-users/topics).

## Checklist

* [x] Does your title concisely summarize the problem?
* [x] Did you include a minimal, reproducible example?
* [x] What OS are you using?
* [x] What version of Dramatiq are you using?
* [x] What did you do?
* [x] What did you expect would happen?
* [x] What happened?

Creating two func decorated with `@actor` without setting its `actor_name` in different module, import the first registered func and send message to run, and you may find that dramatiq run the second but not the first func...

It seems dramatiq recognize func defaultly by its func name...

In many cases, the function names in different modules are set to the same. I think that it is an unreasonable design to use the function name as the only index.

https://github.com/Bogdanp/dramatiq/blob/4b6ad185d4656b4ea7109cd38ef3f40ad17e0566/dramatiq/actor.py#L258

Should we change the default `actor_name` to `actor_name = actor_name or f'{fn.__module__}.{fn.__name__}'`?

I am not familiar with the source code. Based on my initial understanding, it seems that actor_name is only used as the key of the destination function and can be set at will without duplication.

spumer:
I think we need to raise error if try to register actor with same name and require change it.

And to auto resolve name conflict is best to use `__qualname__` https://devdocs.io/python~3.12/reference/datamodel#function.__qualname__

`__qualname__` include full path to func

z0z0r4:
maybe I will have a PR recently

shshe:
We've encountered this problem as well. We added a Dramatiq middleware to detect duplicate actor names:

```python
class EnforceUniqueActorNames(Middleware):
    def before_declare_actor(self, broker, actor):
        if actor.actor_name in broker.actors:
            logger.error(
                "Duplicate actor name detected: %s. Actor names must be unique. "
                "Set a unique name using the `actor_name` kwarg or by renaming the "
                "function decorated by @dramatiq.actor.",
                actor.actor_name,
            )
            raise DuplicateActorError
```

Wealso  use a replacement for the standard `@actor` decorator that registers the actors with the module name prefix (aligning it with the same behavior as Celery):


```python
def actor_v2(
    fn: Optional[Callable] = None,
    *,
    actor_name: Optional[str] = None,
    **options,
) -> Union[Actor, Callable]:

    def decorator(fn):
[...]

PR #641, opened 2024-07-16
feat: update actor_name assignment in decorator

z0z0r4:
Close https://github.com/Bogdanp/dramatiq/issues/640

### DIFF
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -258,7 +258,7 @@ def actor(
     """
     def decorator(fn: Callable[..., Union[Awaitable[R], R]]) -> Actor[P, R]:
         nonlocal actor_name, broker
-        actor_name = actor_name or fn.__qualname__
+        actor_name = actor_name or fn.__name__
         if not _queue_name_re.fullmatch(queue_name):
             raise ValueError(
                 "Queue names must start with a letter or an underscore followed "
--- a/dramatiq/actor.py
+++ b/dramatiq/actor.py
@@ -64,6 +64,9 @@ class Actor(Generic[P, R]):
         priority: int,
         options: Dict[str, Any],
     ) -> None:
+        if actor_name in broker.actors:
+            raise ValueError(f"An actor named {actor_name!r} is already registered.")
+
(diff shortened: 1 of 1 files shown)


Bogdanp:
This isn't the right approach, since it would be a breaking change to any currently-deployed code. I would prefer it if we checked for duplicate actor names and raised an error instead.

z0z0r4:
Indeed, I think checking for duplicate actor_name is necessary. But at the same time, it is reasonable to change the default value. We should support functions decorated with the same function name by default, right? This should also be a design mistake.

Bogdanp:
I think it's more nuanced than a plain design mistake. At least, I remember thinking about this quite a bit when I made the original decision to use `__name__`. For example, even if we used `__qualname__` here, renaming the module in which an actor is defined would then potentially lead to a production bug (actors are looked up by name, and messages already enqueued for `mod1.foo` would not be able to find `mod2.foo`). Using `__qualname__` now is out of the question since it would break users' code for the same reason. So, the only good option is to keep using `__name__` and to error when there are duplicates.

z0z0r4:
> actors are looked up by name, and messages already enqueued for mod1.foo would not be able to find mod2.foo

Will the production environment keep the queue unemptied and update dramatiq? I don't think so, but there is no evidence.

I will keep this pr until I open another to add the exception you said.

z0z0r4:
@Bogdanp fixed

dramatiq/actor.py
```diff
@@ -255,7 +258,7 @@ def actor(
     """
     def decorator(fn: Callable[..., Union[Awaitable[R], R]]) -> Actor[P, R]:
         nonlocal actor_name, broker
-        actor_name = actor_name or fn.__name__
+        actor_name = actor_name or fn.__qualname__
```
Bogdanp, line 261: This change shouldn't be here, no?

Bogdanp (changes requested):

dramatiq/actor.py
z0z0r4: yeee, sorry

z0z0r4:
fixed

PR #645, opened 2024-07-23
Update Celery supported features (Motivation/Compared to * update)

smiling-watermelon:
Some changes here are debatable, and, perhaps require additional review.

For example:
`py2_cel` - In the [documentation](https://docs.celeryq.dev/en/stable/contributing.html#maintenance-branches) it's not clear weather 4.2 here refers to the version of celery, or not. Because if it's maintained, then Python 2 is not dropped to this day, but this line was unchanged for 6 years now (according to [Blame](https://github.com/celery/celery/blame/main/CONTRIBUTING.rst#L347)), and there were no releases to 4.*.* series of Celery since 2020.

`win` - I guess leaving it at "No" is also correct, but Celery `may`, or `may not` work on Windows.

`cel_rate` - Support for rate limits is present all the way to version 3.1 (didn't look any further), so, perhaps it's stated as "No" because this is an extension.

I would suggest a re-evaluation for all libraries in the table, though, and re-check Celery too.
Giving away this contribution in hopes to push the metaphorical stone on the matter of outdated comparisons.

### DIFF
--- a/docs/source/motivation.rst
+++ b/docs/source/motivation.rst
@@ -34,57 +34,57 @@ I've used Celery_ professionally for years and my growing frustration
 with it is one of the reasons why I developed dramatiq.  Here are some
 of the main differences between Dramatiq, Celery, Huey and RQ:
 
-+------------------------------+----------+---------------+--------------+--------------+
-|                              | Dramatiq | Celery_       | Huey_        | RQ_          |
-+------------------------------+----------+---------------+--------------+--------------+
-| Python 2 support             | No       | Yes           | Yes          | Yes          |
-|                              |          |               |              |              |
-+------------------------------+----------+---------------+--------------+--------------+
-| Windows support              | Yes      | No            | Yes          | No           |
-|                              |          |               |              |              |
-+------------------------------+----------+---------------+--------------+--------------+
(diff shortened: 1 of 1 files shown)


docs/source/motivation.rst
```diff
@@ -34,57 +34,57 @@ I've used Celery_ professionally for years and my growing frustration
...
+| Code auto-reload             | Yes      | No              | No           | No           |
+|                              |          |                 |              |              |
++------------------------------+----------+-----------------+--------------+--------------+
+| RabbitMQ support             | Yes      | Yes             | Yes          | No           |
+|                              |          |                 |              |              |
++------------------------------+----------+-----------------+--------------+--------------+
+| Redis support                | Yes      | Yes             | Yes          | Yes          |
+|                              |          |                 |              |              |
++------------------------------+----------+-----------------+--------------+--------------+
+| In-memory broker support     | Yes      | No              | Yes          | No           |
```
spumer, line 82: Kombu support `:memory:` broker and Celery support it as well

pickfire: I believe we can add dashboard support here too.

PR #646, opened 2024-08-22
Fix bug so `max_retries` can be overridden to zero in message options

chrisittner:
Hi 👋, 

small bugfix for the following issue that I stumbled over:

```python
@dramatiq.actor(max_retries=99)
def do_work():
    raise Exception("attempt failed")

do_work.send_with_options(max_retries=0)
```

-> `do_work` will be retried 100 times instead of once, even though message options should override actor options.


Bogdanp:
Thanks! Can you fix the lint issue when you have a moment? I'll go ahead and merge after.

PR #647, opened 2024-08-27
avoid infinite recursion on unkown attribute access in GenericActors.

### LINKED ISSUE #639: docs: class based actor is empty in the docs
> Documentation bug, so most of the options does not apply

The Class-based Actors section in the documentation https://dramatiq.io/reference.html#class-based-actors is empty: 
image

CaselIT:
This was triggered by sphinx autodoc use of inspect.signature that tried to access the missing `__signature__` attribute on GenericActors, preventing the presence of the documentation for generic actors in the website

Fixes #639


Bogdanp:
Thanks!

PR #659, opened 2024-10-25
Fix #656 -- Add Python 3.13 support

### LINKED ISSUE #656: TimeLimit middleware error on Python 3.13
# Issues

GitHub issues are for bugs.  If you have questions, please ask them on the [mailing list](https://groups.io/g/dramatiq-users/topics).

## Checklist

* [x] What OS are you using?
* [x] What did you do?
* [x] What happened?

## What OS are you using?

I've tested it on MacOS 15 and Debian 12

1.17.0

## What did you do?

Upgraded to Python 3.13 and ran my Dramatiq worker configured with https://github.com/Bogdanp/django_dramatiq.

I expect no errors to be raised when the `"dramatiq.middleware.TimeLimit",` is enabled.

## What happened?

The following error is raised:
```
TypeError: '_thread._ThreadHandle' object is not callable
[2024-10-08 12:48:06,729] [PID 31240] [Thread-1] [dramatiq.middleware.time_limit.TimeLimit] [ERROR] Unhandled error while running the time limit handler.
  File "/Users/anzepecar/code/fedidevs/.venv/lib/python3.13/site-packages/dramatiq/middleware/time_limit.py", line 113, in run
    self._handle()
    ~~~~~~~~~~~~^^
```
I assume that Python 3.13 changed the `_thred._ThreadHandle` signature.

amureki:
`threading.Thread` received a new `_handle` attribute in Python 3.13, which clashes with dramatiq's `_CtypesTimeoutManager(Thread)` class. This commit renames the dramatiq method to resolve the conflict.

It also adds Python 3.13 to CI and package classification.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -8,7 +8,7 @@ jobs:
       - uses: actions/checkout@master
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.12"
+          python-version: "3.13"
       - run: |
           sudo apt-get update
           sudo apt-get remove libhashkit2 libmemcached11 || true
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-20.04"]
-        python: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
         concurrency: ["cpython", "gevent"]
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


Bogdanp:
Thanks!

wreczek:
shouldn't it also be in tox.ini?

amureki:
> shouldn't it also be in tox.ini?

Hm, interesting catch!
From what I see, [tox is only used for docs and linting](https://github.com/search?q=repo%3ABogdanp%2Fdramatiq%20tox&type=code) (on CI) right now:
CleanShot 2024-11-04 at 08 55 45@2x

And the GitHub CI matrix is actually running Python tests.

So, tox Python environments are serving local testing.
If this is a valid and expected use case, I will provide a patch; otherwise it might be due for a clean-up.

@Bogdanp what would you say?

Bogdanp:
I think it's probably fine to keep the tox config, even if it's not used in CI. I do still use it locally to test things every now and then.

PR #669, opened 2024-12-20
fix: make sure RabbitMQ broker’s enqueue() attempts the correct number of preset retries

### LINKED ISSUE #668: RabbitMQ Broker off-by-one error when reattempting ensuring a queue
I think the issue with the [`RabbitmqBroker._ensure_queue()`](https://github.com/Bogdanp/dramatiq/blob/4dd0d7125d33afe660405335b8a8de3287b2bec1/dramatiq/brokers/rabbitmq.py#L248C9-L272) function can be illustrated by this equivalent snippet:
```python
>>> MAX_ATTEMPTS = 2
>>> 
>>> def do_attempts():
...     attempts = 1
...     while True:
...         try:  
...             raise ValueError()
...         except ValueError:
...             attempts += 1
...             if attempts > MAX_ATTEMPTS:
...                 return
...             print(f"Retry [{attempts}/{MAX_ATTEMPTS}]")
```
which executes as follows:
```python
>>> do_attempts()
Retry [2/2]
```
I think the fix is simple: move the increment from the beginning to the end of the `except`, which results in:
```python
>>> do_attempts()
Retry [1/2]
Retry [2/2]
```
Alternatively, start with `attempts = 0` and increment right at the beginning of the loop. A similar pattern exists in the `RabbitmqBroker.enqueue()` function [here](https://github.com/Bogdanp/dramatiq/blob/4dd0d7125d33afe660405335b8a8de3287b2bec1/dramatiq/brokers/rabbitmq.py#L324-L361).

@Bogdanp this is a simple fix, happy to provide a PR.

jenstroeger:
Closes #668

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -246,7 +246,7 @@ class RabbitmqBroker(Broker):
             self._ensure_queue(queue_name)
 
     def _ensure_queue(self, queue_name):
-        attempts = 1
+        attempts = 0
         while True:
             try:
                 if queue_name in self.queues_pending:
@@ -321,7 +321,7 @@ class RabbitmqBroker(Broker):
                 },
             )
 
-        attempts = 1
+        attempts = 0
         while True:
             try:
                 self.declare_queue(canonical_queue_name, ensure=True)


dramatiq/brokers/rabbitmq.py
```diff
@@ -246,8 +246,9 @@ def declare_queue(self, queue_name, *, ensure=False):
             self._ensure_queue(queue_name)
 
     def _ensure_queue(self, queue_name):
-        attempts = 1
+        attempts = 0
```
Bogdanp, line 249: It seems like just changing the initialization to `0` would be enough, no? I don't think we have to move the increment to the top of the loop, but maybe I'm missing something. Ditto for the other snippet.

jenstroeger: I agree, but thought it’s easier to read with incrementing at the beginning of the loop instead of the error only. However, that also _adds_ an unnecessary increment in the success case.

I’m easy either way. Want me to move the increment back down?

Bogdanp: Yeah, let's move it back down. I like having it co-located with the check. Thanks!

Bogdanp:
Thanks again!

jenstroeger:
@Bogdanp I stumbled upon this minor bug when I was investigating this: [Issue with RabbitMQ Broker: sending a message to the queue creates a consumer thread locally](https://groups.io/g/dramatiq-users/message/314).

Could you please take a look? I can open a PR with the proposed solution which works locally, but somehow I’m not too sure if that’s a good solution, it feels a little awkward.

PR #670, opened 2024-12-23
feat: add ensure_consumer kwarg to before/after_declare_queue to control whether a consumer should be created on a RabbitMQ, then disable consumer thread creation upon message delivery

jenstroeger:
Following the discussion [Issue with RabbitMQ Broker: sending a message to the queue creates a consumer thread locally](https://groups.io/g/dramatiq-users/topic/110212159), this draft illustrates the proposed change.

@Bogdanp I also noticed that there’s no
```python
self.emit_before("declare_delay_queue", delayed_name)
```
callback when the DQ is declared — is that intentional?

### DIFF
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -219,7 +219,7 @@ class RabbitmqBroker(Broker):
         self.declare_queue(queue_name, ensure=True)
         return self.consumer_class(self.parameters, queue_name, prefetch, timeout)
 
-    def declare_queue(self, queue_name, *, ensure=False):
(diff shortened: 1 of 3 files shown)


Bogdanp:
Thanks, but, unfortunately, this would be a breaking change for anyone already implementing one of these hooks, so I don't think this can work (and I don't think the problem warrants complicating the code to add arity checks to work around the breakage). Until we can come up with a better solution, I think the way to go is that I proposed in that discussion. A potential alternative would be to allow some level of worker configuration -- eg. to allow a user to extend the worker middleware in some way.

Issue #392, opened 2021-04-07
How to disable prefetching


hmoravec:
There [is implemented](https://github.com/Bogdanp/dramatiq/commit/8a103e087d88835f4b256134036ce5149c230db9#diff-7bc0fe5a9d87fcca70f02f7a623d50329ca9047b0bafd4d40139d40ceb8886fd) possibility to set number of prefetched messages using `dramatiq_queue_prefetch` env var (see #183), however it is not documented anywhere and when it is set to zero, it is ignored because `self.queue_prefetch = QUEUE_PREFETCH or min(worker_threads * 2, 65535)` evaluates the first condition as false. Not sure if this is a bug or a feature. Is it possible to disable prefetching so it does no happen that one worker prefetches a message and the other workers cannot work on it even when they are idle?

Bogdanp:
See the note above the constant's definition:

https://github.com/Bogdanp/dramatiq/blob/cdb5035f4a1ed8ae8159589eee441f06a765983a/dramatiq/worker.py#L41-L45

The value includes in-progress messages so to only fetch a single message at a time, you should set `dramatiq_queue_prefetch` to the number of worker threads you run. If you only have a single worker thread, then setting it to 1 should do what you want.

hmoravec:
I see, in the mentioned PR there was not this comment. 
It would be worth to document this (and other settings via env vars) in the documentation.

OlegWock:
@Bogdanp sorry to raise old issue, but I still can't wrap this around my head. I have 8 processes with 50 (gevent) workers in each of them. If I set `dramatiq_queue_prefetch` to one, will this result in 'handling one message per process at same time' or 'handling 50 messages per process at same time (1 per thread)'?

Thanks

PR #674, opened 2025-01-17
Fix changelog entry for dramatiq_queue_prefetch

LincolnPuzey:
I believe `dramatiq_queue_prefetch` is per worker process, not per thread.

As per the source code https://github.com/Bogdanp/dramatiq/blob/b372f431ae40ff383a5b450dc37c8e5d5671bf49/dramatiq/worker.py#L41

And this comment https://github.com/Bogdanp/dramatiq/issues/392#issuecomment-815458472

This matches my testing where, when running with 2 threads, I needed to set `dramatiq_queue_prefetch=2` to make each thread process a message, but not to pre-fetch any extra ones.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -575,7 +575,7 @@ Added
 ^^^^^
 
 * `dramatiq_queue_prefetch` environment variable to control the number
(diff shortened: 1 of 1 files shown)


Bogdanp:
Thanks!

Issue #676, opened 2025-01-26
Example of using actors from different files


bagowix:
After reading all the documentation, all the examples that are in the repository and searching the whole internet for examples of implementation, I did not find the example I need.

Unfortunately, everywhere it is about initialization of broker and actors in one file, which is not very suitable for production code.

Here is an example of the implementation I was looking for:

FastAPI

app/core/dramatiq_service.py - broker initialization
app/tasks/foo/tasks.py - multiple actors 
app/tasks/bar/tasks.py - multiple actors
app/main.py

first terminal: uvicorn main:app --host 0.0.0.0 --port 8001 --reload
second terminal: dramatiq core.dramatiq_service

All actors are initialized and run correctly.

FinnLidbetter:
I think such an example is provided by running `dramatiq --help`
Here is there relevant part of the output:
```
...
examples:
  # Run dramatiq workers with actors defined in `./some_module.py`.
  $ dramatiq some_module

  # Run with a broker named "redis_broker" defined in "some_module".
  $ dramatiq some_module:redis_broker

  # Run with a broker named "broker" defined as attribute of "app" in "some_module".
  $ dramatiq some_module:app.broker

  # Run with a callable that sets up a broker.
  $ dramatiq some_module:setup_broker

  # Import extra modules.  Useful when your main module doesn't import
  # all the modules you need.
  $ dramatiq some_module some_other_module
...
```

So for what you want, it should be something like
```
# First terminal
first terminal: uvicorn main:app --host 0.0.0.0 --port 8001 --reload

# Second terminal
dramatiq core.dramatiq_service tasks.bar.tasks tasks.foo.tasks
```
Possibly with some adjustments depending on how your broker is defined in `dramatiq_service.py`.
You can see the details of the implementation here https://github.com/Bogdanp/dramatiq/blob/a978028742f1159391799716be411ffc34688eba/dramatiq/cli.py#L395-L400

I think the Google group is a better place for questions like these. This repo's GitHub issue template reads 

>  GitHub issues are for bugs. If you have questions, please ask them on the [mailing list](https://groups.io/g/dramatiq-users/topics).

Although it seems like GitHub has moved to a new version of specifying Issue Templates, so it looks like the templating is broken.

bagowix:
Hi,
thanks for the information!

I created the issue to supplement the existing examples in the repository, as I spent too much time creating this implementation.

PR #677, opened 2025-01-26
using actors from different files

bagowix:
Example of using actors from different files.

Issue: https://github.com/Bogdanp/dramatiq/issues/676


synweap15:
Hey @bagowix, such addition would be useful. However, since dramatiq is framework-agnostic, FastAPI mentions would have to be removed. Thanks!

PR #679, opened 2025-02-04
Allow for pipelines to ignore error and continue.

J08nY:
It is useful for me to have a pipeline of messages that execute. However, I want to keep executing the pipeline even if the individual message fails. This pull-request implements that using a new option on the message in the Pipelines middleware.

### DIFF
--- a/docs/source/cookbook.rst
+++ b/docs/source/cookbook.rst
@@ -144,6 +144,16 @@ call |pipeline_get_results|::
   for res in pipe.get_results(block=True):
       ...
 
+If you want the actors in the pipeline to keep executing even if
+a actor in it errors out, use the ``pipe_ignore_exception`` option::
+
+  (
+      throws_exception.message(pipe_ignore_exception=True) |
+      will_still_run.message()
+  )
+
+Note that if the ``may_fail`` actor throws an exception,
+the ``will_still_run`` actor will receive ``None`` as input.
 
 Error Reporting
 ---------------
--- a/dramatiq/middleware/pipelines.py
+++ b/dramatiq/middleware/pipelines.py
@@ -25,6 +25,8 @@ class Pipelines(Middleware):
     Parameters:
       pipe_ignore(bool): When True, ignores the result of the previous
         actor in the pipeline.
+      pipe_ignore_exception(bool): When True, ignores the fact that the previous
+        actor in the pipeline threw an exception and runs anyway.
       pipe_target(dict): A message representing the actor the current
         result should be fed into.
     """
@@ -33,6 +35,7 @@ class Pipelines(Middleware):
     def actor_options(self):
         return {
             "pipe_ignore",
+            "pipe_ignore_exception",
             "pipe_target",
         }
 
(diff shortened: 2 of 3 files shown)


docs/source/cookbook.rst
```diff
@@ -144,6 +144,16 @@ call |pipeline_get_results|::
   for res in pipe.get_results(block=True):
       ...
 
+If you want the actors in the pipeline to keep executing even if
+a actor in it errors out, use the ``pipe_errorok`` option::
```
synweap15, line 148: @J08nY `pipe_ignore_exception` would be the preferred variable name

J08nY: Renamed now.

dramatiq/middleware/pipelines.py
```diff
@@ -42,10 +45,12 @@ def after_process_message(self, broker, message, *, result=None, exception=None)
         # from broker -> pipelines -> messages -> broker.
         from ..message import Message
 
-        if exception is not None or message.failed:
+        actor = broker.get_actor(message.actor_name)
+
+        errorok = message.options.get("pipe_ignore_exception") or actor.options.get("pipe_ignore_exception")
```
synweap15, line 50: @J08nY I'd rename the left-side to `ignore_exception`

dramatiq/middleware/pipelines.py
```diff
@@ -42,10 +45,12 @@ def after_process_message(self, broker, message, *, result=None, exception=None)
...
+        if (exception is not None or message.failed) and not errorok:
```
synweap15, line 51: @J08nY `exception is not None` and `message.failed` shouldn't be considered equal - they may have different meanings in the actor implementation. Only explicitly thrown exceptions should be caught.

I'd also go for explicit `.. and ignore_exception is not True`

tests/test_composition.py
```diff
@@ -298,6 +298,36 @@ def should_never_run():
...
+    def do_nothing():
+        pass
+
+    @dramatiq.actor
+    def should_run():
+        nonlocal has_run
+        has_run = True
+
+    # When I pipe some messages intended for that actor together and run the pipeline
+    pipe = do_nothing.message_with_options(pipe_ignore_exception=True) | should_run.message_with_options(pipe_ignore=True)
```
synweap15, line 321: @J08nY with the above, the changes would require a redo

dramatiq/middleware/pipelines.py
J08nY: That is why I had the name "errorok", to consider both cases of message failures. This can of course be split, but that requires a new parameter.

LincolnPuzey:
@synweap15 My apologies, I didn't know.

Do you agree that continuing a pipeline multiple times due to retries, is something we should avoid?

J08nY:
Are there some semantics of this concept that you want to ignore exceptions in an actor in a pipeline, that work together with retries?

The only thing I can think of is to make the actor eat the exception and thus effectively disable retries.

J08nY:
Any updates on this? I would still like to be able to have pipelines with actors that fail, yet the pipeline continues.

PR #685, opened 2025-03-11
fix: message can be covariant in result type

### LINKED ISSUE #684: Message can be covariant in the result type
Currently, this code runs fine, but doesn't type check:

```python
class Superclass:
    pass

class Subclass(Superclass):
    pass


@actor
def actor_1() -> Superclass:
    return Superclass()

@actor
def actor_2() -> Subclass:
    return Subclass()


task: Message[Superclass]

if use_subclass:
    task = actor_2.message()  # error: cannot assign Message[Subclass] to Message[Superclass]
else:
    task = actor_1.message()

result: Superclass = task.get_result()
```

This can be fixed by making `Message[R]` covariant in `R`. I've sent a PR which makes this change.

alecbarber:
resolves #684

### DIFF
--- a/dramatiq/message.py
+++ b/dramatiq/message.py
@@ -29,7 +29,7 @@ from .results import ResultBackend
 #: The global encoder instance.
 global_encoder: Encoder = JSONEncoder()
 
-R = TypeVar("R")
+R = TypeVar("R", covariant=True)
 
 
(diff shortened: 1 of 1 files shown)


Bogdanp:
Thanks!

PR #687, opened 2025-03-17
Fix flushing of RabbitMQ queues

olii:
This PR fixes a bug when queues on the `broker` are flushed.

The old behavior was incorrect. It didn't flush delayed queues ending with `.DQ` and `.XQ` because the main queue was in the `self.queues_pending` variable (e.g. undeclared yet). We will have to ensure queues exist by declaring them first.

The new behavior doesn't require queues to be declared. If the purge fails and the RabbitMQ server closes the channel, it is recreated again, and the flushing continues.

This PR doesn't come with tests. I was not able to find them for the existing code. Based on that I didn't initiate the discussion as this is only useful in dev/tests environment.


Bogdanp:
Thanks!

PR #688, opened 2025-03-27
make Prometheus optional

### LINKED ISSUE #345: Remove Prometheus from the list of default middleware
# Issues

Currently (1.9.0) the prometheus middleware is among the default middleware that are used if an user does not specify any.
Also by default it connects to the `0.0.0.0` address, making the server accessible on the external network.
https://github.com/Bogdanp/dramatiq/blob/13e8d626c4d0250a92a436324b0039204a3cf706/dramatiq/middleware/prometheus.py#L37

I think it may make sense to remove it from the default middleware, since it offers functionality that is not likely to be used by the majority of the users.

At least it should only connect to `localhost` by default and not to `0.0.0.0`.

## What OS are you using?

windows (not relevant here

1.9.0

prometheus is an optional middleware

## What happened?

prometheus is strarted by default and exposes information on 0.0.0.0

azmeuk:
I applied the patch proposed in #345 to make the Prometheus middleware optional:

- prometheus-client is now an optional dependency
- The Prometheus middleware is only loaded if the prometheus_client is installed.

fixes #345

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -38,9 +38,9 @@ Changed
   (`#640`_, `#641`_, `@z0z0r4`_)
 * The `Retries` middleware now tracks when a message was last requeued
   on retry. (`#629`_, `@kuba-lilz`_)
-* The `prometheus-client` dependency is optional.
-  The `Prometheus` middleware is only loaded when the `prometheus-client`
-  dependency is installed. (`#345`_, `#688`_, `@azmeuk`_)
+* The `prometheus-client` dependency is now optional, and the `Prometheus`
+  middleware is no longer in the default list, and should be added manually.
+   (`#345`_, `#688`_, `@azmeuk`_)
 
 .. _#345: https://github.com/Bogdanp/dramatiq/issues/345
 .. _#629: https://github.com/Bogdanp/dramatiq/pull/629
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -38,10 +38,16 @@ Changed
   (`#640`_, `#641`_, `@z0z0r4`_)
 * The `Retries` middleware now tracks when a message was last requeued
   on retry. (`#629`_, `@kuba-lilz`_)
+* The `prometheus-client` dependency is optional.
+  The `Prometheus` middleware is only loaded when the `prometheus-client`
+  dependency is installed. (`#345`_, `#688`_, `@azmeuk`_)
 
+.. _#345: https://github.com/Bogdanp/dramatiq/issues/345
 .. _#629: https://github.com/Bogdanp/dramatiq/pull/629
 .. _#640: https://github.com/Bogdanp/dramatiq/issues/640
 .. _#641: https://github.com/Bogdanp/dramatiq/pull/641
+.. _#688: https://github.com/Bogdanp/dramatiq/pull/688
+.. _@azmeuk: https://github.com/azmeuk
 .. _@kuba-lilz: https://github.com/kuba-lilz
 .. _@z0z0r4: https://github.com/z0z0r4
 
--- a/dramatiq/middleware/__init__.py
+++ b/dramatiq/middleware/__init__.py
@@ -15,8 +15,6 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import importlib
-
 from .age_limit import AgeLimit
 from .asyncio import AsyncIO
 from .callbacks import Callbacks
@@ -24,7 +22,6 @@ from .current_message import CurrentMessage
 from .group_callbacks import GroupCallbacks
 from .middleware import Middleware, MiddlewareError, SkipMessage
 from .pipelines import Pipelines
-from .prometheus import Prometheus
 from .retries import Retries
 from .shutdown import Shutdown, ShutdownNotifications
 from .threading import Interrupt, raise_thread_exception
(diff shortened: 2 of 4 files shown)


dramatiq/middleware/__init__.py
```diff
@@ -44,5 +46,12 @@
 
 #: The list of middleware that are enabled by default.
 default_middleware = [
-    Prometheus, AgeLimit, TimeLimit, ShutdownNotifications, Callbacks, Pipelines, Retries
+    AgeLimit, TimeLimit, ShutdownNotifications, Callbacks, Pipelines, Retries
 ]
+
+try:
+    if importlib.util.find_spec("prometheus_client"):
+        default_middleware.append(Prometheus)
```
Bogdanp, line 54: This should insert the middleware at index 0 to preserve the existing behavior.

setup.py
```diff
@@ -39,10 +39,6 @@ def rel(*xs):
         raise RuntimeError("Version marker not found.")
 
 
-dependencies = [
-    "prometheus-client>=0.2",
```
Bogdanp, line 43: This would technically be a breaking change to the build process for people who use Dramatiq, so I'm not sure what to do about it. Is there a way to specify negative dependencies in `setup.py`? Eg., `pip install dramatiq[no-prometheus]`. If so, that's the way we should go.

azmeuk: I am not sure that there is a thing such as negative dependencies. I think this is covered by [PEP771](https://peps.python.org/pep-0771/) but this is still a draft at the moment.

Removing `prometheus-client` is indeed a breaking change, and I am not sure how this could be otherwise. Usually such changes come with deprecation warnings, but in the case of dependencies, I don't know how to raise a deprecation warning for users that use the middleware but have not explicitly added its extra.

The only remaining option I see is good communication on the changelog.

dramatiq/middleware/__init__.py
```diff
@@ -38,11 +40,18 @@
...
 
 #: The list of middleware that are enabled by default.
 default_middleware = [
-    Prometheus, AgeLimit, TimeLimit, ShutdownNotifications, Callbacks, Pipelines, Retries
+    AgeLimit, TimeLimit, ShutdownNotifications, Callbacks, Pipelines, Retries
 ]
+
+try:
+    if importlib.util.find_spec("prometheus_client"):
+        default_middleware.insert(0, Prometheus)
```
LincolnPuzey, line 54: My suggestion: if we are going to make a breaking change release removing Prometheus from the default middleware, we should also remove this line. In case there is someone who wants to install this library, but doesn't want the middleware.

i.e. the `default_middleware` list shouldn't change depending on what libraries you have installed.

azmeuk: @Bogdanp, do you have an opinion about this?

azmeuk:
Hi @frankie567 @oYASo @sfahad1414 @SamR1 @mhdzumair
I summoned you because we would like some advice on your usage of dramatiq, and especially if you have an opinion about removing prometheus as a required dependency.

> I would love it if folks who operate Dramatiq in production would weigh in and let us know what they think.

I summoned you because your projects are the most popular from the [dramatique dependency graph](https://github.com/Bogdanp/dramatiq/network/dependents).

Any thoughts on the matter?

mhdzumair:
You can make it optional.

sfahad1414:
Yes @azmeuk can be made optional.

azmeuk:
Thank you all for your answers!

@Bogdanp do you need more feedback? I can post something on the ML if you want.

Bogdanp:
Yes, let's do a post on the mailing list and if no one objects within a week or so we can go ahead with this. Thanks for your work on this!

guedesfelipe:
Need changes at doc? https://dramatiq.io/advanced.html#prometheus-metrics

karolinepauls:
I get around some of this problem by specifying all middleware explicitly, as a list - maybe the docs should also move this direction. It won't solve the issue of prometheus being added but at least users following the docs will be able to easily remove it.

azmeuk:
@guedesfelipe you mean adding a note to indicate that Prometheus must be installed as an extra?

guedesfelipe:
> @guedesfelipe you mean adding a note to indicate that Prometheus must be installed as an extra?
> 
> @karolinepauls could it is the subject of another PR?

Something like that.

And add an error when importing the prometheus client if it is not installed correctly?

azmeuk:
> And add an error when importing the prometheus client if it is not installed correctly?

@guedesfelipe Isn't the natural `ImportError` enough?

> @azmeuk, how do they feel on the ML?

@KrilleGH the ML has been quiet, so no one has raised objections.

@Bogdanp, what do you think?

Bogdanp (changes requested):
Had one last minor comment on this and I think it can be merged.

> Isn't the natural ImportError enough?

I agree that it is enough.

docs/source/changelog.rst
```diff
@@ -38,10 +38,16 @@ Changed
   (`#640`_, `#641`_, `@z0z0r4`_)
 * The `Retries` middleware now tracks when a message was last requeued
   on retry. (`#629`_, `@kuba-lilz`_)
+* The `prometheus-client` dependency is optional.
+  The `Prometheus` middleware is only loaded when the `prometheus-client`
```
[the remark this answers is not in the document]
azmeuk: Done.

Bogdanp:
Thanks @azmeuk for carrying this through!

PR #695, opened 2025-04-15
Fix TypeError when using decorator on a module-less function

### LINKED ISSUE #694: Runtime error when using `@actor` decorator in interactive shell with a locally defined function
I was debugging some of our own code in an interactive Python shell, and ran into a runtime type error with the `@actor` decorator:

```
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'
```

The error is caused by this line, which fails with locally defined, non module-bound functions:

https://github.com/Bogdanp/dramatiq/blob/6d60a5fdf6755b846781a1adc06dfa062a1186f9/dramatiq/actor.py#L70

It's definitely an edge case, but this simple change fixes it:

```py
self.logger = get_logger(fn.__module__ or '_', actor_name)
```

bartvanandel:
Fixes #694

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -67,3 +67,4 @@ of those changes to CLEARTYPE SRL.
 | [@kuba-lilz](https://github.com/kuba-lilz/)            | Jakub Kolodziejczyk    |
 | [@dbowring](https://github.com/dbowring/)              | Daniel Bowring         |
 | [@olii](https://github.com/olii)                       | Oliver Nemček          |
+| [@bartvanandel](https://github.com/bartvanandel)       | Bart van Andel         |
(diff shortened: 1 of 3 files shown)


synweap15:
@bartvanandel it's now behind master, please rebase on top of it.

bartvanandel:
> it's now behind master, please rebase on top of it.

@synweap15 Done.

bartvanandel:
@Bogdanp The failed unit test has nothing to do with my changes. Some timeout occurred.

synweap15:
@bartvanandel indeed and it's reoccuring, I'll make amendments on top of #697 first to ensure tests pass and then pivot back to you here

synweap15:
@bartvanandel Thanks!

PR #697, opened 2025-04-16
Update CI to use ubuntu 24.04

LincolnPuzey:
Ubuntu 20.04 has been removed as a runner, see https://github.com/actions/runner-images/issues/11101


synweap15:
This is an urgent one. The failing test looks flaky, can you retry?

PR #699, opened 2025-04-22
fix: join_queue implementation for compatibility with gevent 25.4.1

synweap15:
Fixes compatibility issues with gevent 25.4.1, released 17th April 2025. See https://github.com/gevent/gevent/issues/2099

The issue caused dramatiq's gevent test suite to fail.


synweap15:
I've added a thread name, please merge it if you approve.

LincolnPuzey:
Looks reasonable to me. At least in dramatiq's own test suite, this is called with `timeout=None` almost all of the time, which will use the simple path.

Issue #701, opened 2025-04-29
Can't start Prometheus middleware in prod container


dlip:
It works fine locally, but in prod _run_exposition_server is never called. after_process_boot is called though.

I have tried running with -p 1 -t 1 and --use-gevent but it didn't help.

dlip:
I've dug into the code a bit more:

The call from cli.py to canteen_get to get the list of middleware to fork results in an empty list because it calls the wait function which determines canteen.initialized = False

The confusing part is canteen_try_init is called and cv.initialized = True. My guess is there is something wrong in `with cv.get_lock()` which is part of contextmanager, not modifying the object correctly

I can hack it to work by changing canteen_get to `return ['dramatiq.middleware.prometheus:_run_exposition_server']` which I think proves this context has an issue

guedesfelipe:
I think I found my problem @dlip, it's because of this line: https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/cli.py#L522... Our projects take more than 30 seconds to start up completely. And this explains why when I increase the CPU of the pods in some circumstances it starts working again.

Maybe it would be interesting to pass a parameter for this timeout and not have it be something fixed, what do you think?

@karolinepauls @synweap15 @Bogdanp (Srry for the ping 🤍)

synweap15:
Hey @guedesfelipe, thanks for the help with debugging this issue. I think it would be good to have the timeout value parametrized, with the default as it is currently - 30 seconds.

What do you think @Bogdanp? I could take on it.

dlip:
@guedesfelipe Unfortunately that doesn't seem to be the issue for me, even setting it to 120s didn't work. I don't think my app takes 30s to startup, especially on a single thread/process.

I also tried setting multiprocessing.set_start_method to spawn, fork and forkserver without success

guedesfelipe:
What a pity @dlip , can you share your code to reproduce it and we can help you with this?

dlip:
@guedesfelipe unfortunately not, its proprietary work code. I'll probably just use this ugly hack until I find more time to test it, sorry if it gives you nightmares 😆 

```
DRAMATIQ_FILE="/usr/local/lib/python3.12/site-packages/dramatiq/canteen.py"
if [[ -f "${DRAMATIQ_FILE}" ]]; then
[...]

PR #706, opened 2025-05-08
Fix/worker startup timeout

guedesfelipe:
#701

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -69,3 +69,4 @@ of those changes to CLEARTYPE SRL.
 | [@olii](https://github.com/olii)                       | Oliver Nemček          |
 | [@bartvanandel](https://github.com/bartvanandel)       | Bart van Andel         |
 | [@LincolnPuzey](https://github.com/LincolnPuzey)       | Lincoln Puzey          |
+| [@guedesfelipe](https://github.com/guedesfelipe)       | Felipe Guedes          |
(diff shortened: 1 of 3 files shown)


synweap15:
hey @guedesfelipe, the [tests are now timing out](https://github.com/Bogdanp/dramatiq/actions/runs/14910226999/job/42056139805?pr=706). I have confirmed the behavior in my local environment and lack of this behavior on current `master`.

Github Actor logging outputs only completed lines, but the issue is in `test_cli.py`, `test_cli_fork_functions_have_no_blocked_signals` test.

image

guedesfelipe:
What is that param `--fork`? https://github.com/Bogdanp/dramatiq/blob/master/tests/test_cli.py#L258

guedesfelipe:
I think the error happens because of the parameter name starting with `fork`, I changed it to `timeout-fork` and the test passed:

image

@synweap15 Do you have a better name suggestion for the parameter?

synweap15:
Indeed it caused the parameter name to be ambigious.

I suggest that:
- the parameter gets renamed to `--worker-fork-timeout`
- move it a few lines up, so it appears above `--worker-shutdown-timeout` argument definition - so then it's in logical order of fork function -> worker-fork-timeout -> worker-shutdown timeout
- I think I would change the value from seconds to miliseconds - the use of miliseconds is consistent throughout the codebase.

synweap15:
@guedesfelipe could you please address the lint fail?

synweap15:
@guedesfelipe looking good, thanks!

guedesfelipe:
Great @synweap15! Do you have any idea when the next release will be out?

PR #707, opened 2025-05-12
Close write_pipe after forking

dansimko:
The write_pipe never goes out of scope and so the last process's write_pipe is never closed by the watcher process, which results in a hang on read() from this pipe in log_watcher when the worker process is killed, and blocking all worker processes when their respective write pipe buffers are exhausted.

### DIFF
--- a/dramatiq/cli.py
+++ b/dramatiq/cli.py
@@ -513,6 +513,7 @@ def main(args=None):  # noqa
         worker_pipes.append(read_pipe)
         worker_processes.append(proc)
         worker_process_events.append(event)
+        write_pipe.close()
 
     # Wait for all worker processes to come online before starting the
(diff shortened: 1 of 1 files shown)


thrix:
Looks like a common issue with Python multiprocessing:

```
Before Fix:
Parent: [read_pipe, write_pipe]   ❌ <- still holds write_pipe
Child:              [write_pipe]
EOF never happens!

After Fix:
Parent: [read_pipe] ✅
Child:              [write_pipe]
→ EOF is detected correctly when child closes or exits.
```

Interestingly, we do not hit it always, but this was hunting our production for quite some time now :) nice find @dansimko

synweap15:
@dansimko great find. Thanks!

Issue #708, opened 2025-05-23
Unexpected Multiple Executions in Dramatiq Pipeline


optimus-kart:
## Checklist

* [ x ] Does your title concisely summarize the problem?
* [ x ] Did you include a minimal, reproducible example?
* [ x ] What OS are you using?
* [ x ] What version of Dramatiq are you using?
* [ x ] What version of Python are you using?
* [ x ] What did you do?
* [ x ] What did you expect would happen?
* [ x ] What happened?

## What OS are you using?

 macOS 15.4.1  

## What version of Dramatiq are you using?
1.17.1

## What version of Python are you using?

Python 3.11.11

## What did you do?

I have the following tasks
```
@dramatiq.actor
def task1(inp:str)->str:
    logger.info("Running task1")
    return "some_data"


@dramatiq.actor
def task2(file_data:str)->None:
    logger.info("Running task2")
    pass

@dramatiq.actor
def task3(arg1:str, arg2:str)->List[str]:
   logger.info("Running task3")
    return ["res1", "res2"]

@dramatiq.actor(store_results=True)
def task4(arg:List[str])->str:
    logger.info("Running task4")
    return "here is your answer"
```

I am running this via pipeline
```
        pipe = pipeline(
            [
                task1.message(arg1),
                task2.message(),
                task3.message_with_options(
                    args=(
                        arg2,
                        arg3,
                    ),
                    pipe_ignore=True,
                ),
                task4.message(),
            ]
        )
```

On running this pipeline I see the following o/p
```
running task1
running task2
running task2
running task3
running task3
running task3
running task3
running task4
running task4
running task4
running task4
running task4
running task4
running task4
running task4
```

I see that each task is executed 2^x times (x being the order when the task is run).

## What did you expect would happen?

Based on the documentation I expected it to run each task only once by piping the o/p of one task to the next.

## What happened?

Tasks are being executed more than the expected number of times

LincolnPuzey:
I suspect you have the `Pipelines` middleware enabled twice, as that would mean the next task in the pipeline is started twice each time.


The `Pipelines` middleware is included in the default middleware, and if you are adding it manually, that would make it be enabled twice.

Can you share any place you are calling `.add_middleware(...)` and/or your list of middleware passed to the broker (if not using the default one?)

optimus-kart:
[...]

PR #709, opened 2025-05-27
feature: log warning when middleware of a given type is added twice

synweap15:
See: https://github.com/Bogdanp/dramatiq/issues/708

### DIFF
--- a/dramatiq/broker.py
+++ b/dramatiq/broker.py
@@ -123,6 +123,9 @@ class Broker:
         You can specify another middleware (by class) as a reference
         point for where the new middleware should be added.
 
+        If there's already a middleware object of the same class
+        added to the broker, it will be replaced.
+
         Parameters:
           middleware(Middleware): The middleware.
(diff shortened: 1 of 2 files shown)


dramatiq/broker.py
```diff
@@ -123,6 +123,9 @@ def add_middleware(self, middleware, *, before=None, after=None):
         You can specify another middleware (by class) as a reference
         point for where the new middleware should be added.
 
+        If there's already a middleware object of the same class
+        added to the broker, it will be replaced.
```
karolinepauls, line 127: Is this comment correct? The test below asserts that both middlewares of the same type are present ("I expect both middlewares to be added").

synweap15: @karolinepauls absolutely, I was testing the described approach on the side and I forgot to replace that comment. Update pushed

LincolnPuzey (approved):
Looks good

karolinepauls (approved):

PR #721, opened 2025-06-20
Change `compute_backoff` implementation to respect min_backoff

### LINKED ISSUE #651: For first retry `compute_backoff` always return value below `min_backoff` settings
## Checklist

* [x] What OS are you using?
* [x] What version of Dramatiq are you using? - 1.17.0
* [x] What did you do?
* [x] What happened?

Dramatiq calculate delay time like this:

```python
            min_backoff = message.options.get("min_backoff", actor.options.get("min_backoff", self.min_backoff))
            max_backoff = message.options.get("max_backoff", actor.options.get("max_backoff", self.max_backoff))
            max_backoff = min(max_backoff, DEFAULT_MAX_BACKOFF)
            _, delay = compute_backoff(retries, factor=min_backoff, max_backoff=max_backoff)
```
https://github.com/Bogdanp/dramatiq/blob/3e20bdd6c536f56822ce8e224c003b336d555be2/dramatiq/middleware/retries.py#L130

When message retries first time we have a `retries = 0` in this scope.
Then `compute_backoff` return value always below `min_backoff`
This happend due default `jitter=True` flag in `compute_backoff` function

You can check it manually:

```ipython
from dramatiq.common import compute_backoff
retries = 0
compute_backoff(retries, factor=1000, max_backoff=5000)
Out[4]: (1, 734)
compute_backoff(retries, factor=1000, max_backoff=5000)
Out[5]: (1, 527)
compute_backoff(retries, factor=1000, max_backoff=5000)
Out[6]: (1, 924)
compute_backoff(retries, factor=1000, max_backoff=5000)
Out[7]: (1, 625)
```

## What OS are you using?

macOS 14.4.1

1.17.0

## What did you do?

Run task with expected delay

Delay should be between min_backoff and max_backoff

## What happened?

Calculated delay below min_backoff

LincolnPuzey:
Closes #651

Currently, because the 'jittering' makes the `backoff` smaller, for the first attempt, the returned value will be smaller than `factor` - which is `min_backoff`. This is the bug reported.

My change here is to make the 'jittering' _increase_ the backoff, so it is never smaller than `factor` (`min_backoff`).

This implementation does mean there will be no jittering once max_backoff is reached.
Is that a problem? @Bogdanp Can I get your opinion?

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,6 +8,19 @@ All notable changes to this project will be documented in this file.
 `Unreleased`_
 -------------
 
+Fixed
+^^^^^
+
+* Fixed the way that backoff time is calculated for retries so that the
+  first backoff is not less than ``min_backoff``.
+  This means that each retry backoff is now, on average, twice as long.
+  To avoid this effect, you can halve your ``min_backoff``
+  (which should now be correctly observed).
+  (`#651`_, `#721`_, `@LincolnPuzey`_)
+
+.. _#651: https://github.com/Bogdanp/dramatiq/issues/651
+.. _#721: https://github.com/Bogdanp/dramatiq/pull/721
+
 Changed
 ^^^^^^^
 
--- a/dramatiq/common.py
+++ b/dramatiq/common.py
@@ -41,17 +41,29 @@ def compute_backoff(attempts, *, factor=5, jitter=True, max_backoff=2000, max_ex
     Parameters:
       attempts(int): The number of attempts there have been so far.
       factor(int): The number of milliseconds to multiply each backoff by.
+                   This is also effectively the minimum backoff amount.
       max_backoff(int): The max number of milliseconds to backoff by.
       max_exponent(int): The maximum backoff exponent.
 
     Returns:
       tuple: The new number of attempts and the backoff in milliseconds.
     """
+    # Calculate base backoff from number of attempts.
+    # When attempts=0, then backoff=factor, i.e. the minimum backoff.
+    # each extra attempt, doubles this.
     exponent = min(attempts, max_exponent)
-    backoff = min(factor * 2 ** exponent, max_backoff)
+    backoff = factor * (2 ** exponent)
+
     if jitter:
-        backoff /= 2
-        backoff = int(backoff + uniform(0, backoff))
+        # Add jitter to extend backoff up to twice as much.
+        backoff = int(backoff * uniform(1, 2))
+        # If this exceeds max_backoff, bring backoff back to 50%-100% of max_backoff to keep some jitter.
+        if backoff > max_backoff:
+            backoff = int(max_backoff * uniform(0.5, 1))
+    else:
+        # if not jittering, simply apply max_backoff
+        backoff = min(backoff, max_backoff)
+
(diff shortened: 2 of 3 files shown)


LincolnPuzey:
@Bogdanp 

> I suspect it doesn't matter in practice, but it's hard to say.

I would say that assuming the first several backoffs are jittered, once max_backoff is reached it doesn't matter, since tasks that started their first attempt at the same time, are no longer in sync.

>  This change will cause most tasks to be delayed more

I reckon, probably doesn't matter in practice.
Or we could make this change and halve the `DEFAULT_MIN_BACKOFF`, so anyone using all the defaults, keeps the same behavior.

> An in-between change could be to just change the return to `return attempts + 1, max(factor, backoff)`

That would mean the first backoff is exactly `factor` with no jitter

Bogdanp:
> Or we could make this change and halve the DEFAULT_MIN_BACKOFF, so anyone using all the defaults, keeps the same behavior.

This would still end up affecting people who've explicitly set a min backoff.

> That would mean the first backoff is exactly factor with no jitter

Yes, and IIRC that's why I went with the current approach, but I don't think it would be that bad to not have jitter on the first attempt.

All that said, I am fine with any of these approaches. Ideally, we would come up with a couple workload examples (as in, code in the `examples/` dir) to see if the change is going to have any impact (eg. to throughput and to worker memory consumption), but I understand that would be quite a bit of work.

LincolnPuzey:
To gauge the effect, these are some example number:

To simulate task reties:
```python
for i in range(20):
    print(dramatiq.common.compute_backoff(i, factor=15000, max_backoff=(86400000 * 7)))
```
results:
```txt
Old               New
(1,       7991)   (1,      29091)
(2,      28551)   (2,      54575)
(3,      54695)   (3,      91485)
(4,     110217)   (4,     213304)
(5,     192363)   (5,     417787)
(6,     355626)   (6,     724451)
(7,     801379)   (7,    1409588)
(8,    1721092)   (8,    3666171)
(9,    3771939)   (9,    4821197)
(10,   7478468)   (10,  12933777)
(11,  15234240)   (11,  15529418)
(12,  23771112)   (12,  56590998)
(13,  43314201)   (13,  76367250)
(14,  88593935)   (14, 224890672)
(15, 178710296)   (15, 477088020)
(16, 380335431)   (16, 511192657)
(17, 572914695)   (17, 527386475)
(18, 519998459)   (18, 497887595)
(19, 576937344)   (19, 564550618)
(20, 548646440)   (20, 326744272)
```
New ones are about 2x, as expected. Both hit the max_backoff for the last ~5.

LincolnPuzey:
> Also, consider that the Redis broker and the results backend use compute_backoff

I think there it will be a similar thing, the backoffs will be 2x on average, although the max_backoff it a lot lower, and is reached sooner.

@Bogdanp I could halve the `factor` in those places so that there should be no change in behavior?

Bogdanp (approved):

Bogdanp:
> @Bogdanp I could halve the factor in those places so that there should be no change in behavior?

Let's keep them as is for now. The factors in those other cases are small enough that this probably won't matter.

Issue #521, opened 2023-01-05
FeatReq: Typehints in dramatiq.middleware


dragonpaw:
Just a small request (doesn't everyone say that?). I love the ability to use my own middleware to customize the messages and add some custom logging. One thing I noticed is there's no type type hinting in the dramatiq.middleware.Middleware class. As a new developer to the package, for places where I am meant to extend the code, it's really helpful to have some hints as to the types to expect. Also might be good to note in the docs that you can alter the arguments, such as the message, and that will be propigated.

Simplified example of what I mean:
```python
class LoggingMiddleware(dramatiq.middleware.Middleware):
    def after_enqueue(
        self,
        broker: dramatiq.broker.Broker,
        message: dramatiq.message.Message,
        delay: int | None,
    ):
        logger.debug("Queued job: %r", message)
``` 

It's a small thing sure, but it does make it easier to write middleware.

jenstroeger:
Perhaps this discussion should be expanded to _all_ of the Dramatic package: the current implementation doesn’t seem to use type hints (although I saw the odd use of it).

Perhaps, though, @Bogdanp plans to add them with v2 at some point?

dragonpaw:
Definitely would be nice to do all of it. I asked specifically about the middleware as there's no documentation on what the arguments to the middleware functions actually are (just their names) and I need to write some to deal with Postgres schema changing while running jobs. So I had to do some custom logging to figure out what the arguments were to be able to make a middleware.

dragonpaw:
A little bit of an update with the latest version just out:

We have a number of custom middlewares, and in some of them we will set `message.failed = True` in before_process_message, in order to have the message thrown in the dead letter queue for possible manual review. But the new dataclass of `dramatiq.Message` doesn't have this property. Technically I know it's actually using the MessageProxy,which does have failed, but I wasn't sure if that's the contract we should be writing to in the middlewares.

PR #731, opened 2025-07-07
Improve typing coverage on external API, in particular Worker and Broker classes

### LINKED ISSUE #727: Add type hints for Broker and Worker class
Related:

- https://github.com/Bogdanp/dramatiq/issues/305
- https://github.com/Bogdanp/dramatiq/pull/306
- https://github.com/Bogdanp/dramatiq/issues/575

There’s been quite some discussion on typing in the past, with only little progress. Furthermore, Dramatiq aims to support Python 3.9 so typing support needs to remain somewhat simple.

My personal need would be a typed [Broker](https://github.com/Bogdanp/dramatiq/blob/d527929a9e08305db29ddd651bb651a217ad368b/dramatiq/broker.py#L72) and [Worker](https://github.com/Bogdanp/dramatiq/blob/d527929a9e08305db29ddd651bb651a217ad368b/dramatiq/worker.py#L54) class because those are the ones I use quite often when testing. 

So, I would be willing to contribute a PR to type the methods of those two classes (nothing as elaborate as closed PR #306, to start with). Any interest?

jenstroeger:
Closes #727

Adds a bunch of type annotations to many public functions; in two cases the annotations would require a minor rewrite of code itself.

I’ve not annotated `message` arguments because they can be either of type `Message` or `MessageProxy` (because they’re used interchangeably). I also refrained from importing modules guarded by `if TYPE_CHECKING` but I can add that too and further improve some method signatures. Please let me know what you prefer!

### DIFF
--- a/dramatiq/broker.py
+++ b/dramatiq/broker.py
@@ -15,12 +15,12 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from typing import Optional, cast
+from typing import Any, Optional, cast
 
 from .errors import ActorNotFound
 from .logging import get_logger
 from .middleware import MiddlewareError, default_middleware
-from .results import Results
+from .results import ResultBackend, Results
 
 #: The global broker instance.
 global_broker: Optional["Broker"] = None
@@ -100,7 +100,7 @@ class Broker:
         for m in middleware:
             self.add_middleware(m)
 
-    def emit_before(self, signal, *args, **kwargs):
+    def emit_before(self, signal: str, *args: Any, **kwargs: Any) -> None:
         signal = "before_" + signal
         for middleware in self.middleware:
             try:
@@ -110,7 +110,7 @@ class Broker:
             except Exception:
                 self.logger.critical("Unexpected failure in %s of %r.", signal, middleware, exc_info=True)
 
-    def emit_after(self, signal, *args, **kwargs):
+    def emit_after(self, signal: str, *args: Any, **kwargs: Any) -> None:
         signal = "after_" + signal
         for middleware in reversed(self.middleware):
             try:
@@ -173,10 +173,10 @@ class Broker:
         for queue_name in self.get_declared_delay_queues():
             middleware.after_declare_delay_queue(self, queue_name)
 
-    def close(self):
+    def close(self) -> None:
         """Close this broker and perform any necessary cleanup actions."""
 
-    def consume(self, queue_name, prefetch=1, timeout=30000):  # pragma: no cover
+    def consume(self, queue_name: str, prefetch: int = 1, timeout: int = 30000) -> "Consumer":  # pragma: no cover
         """Get an iterator that consumes messages off of the queue.
 
         Raises:
@@ -204,7 +204,7 @@ class Broker:
         self.actors[actor.actor_name] = actor
         self.emit_after("declare_actor", actor)
 
-    def declare_queue(self, queue_name):  # pragma: no cover
+    def declare_queue(self, queue_name: str) -> None:  # pragma: no cover
         """Declare a queue on this broker.  This method must be
         idempotent.
 
@@ -213,7 +213,7 @@ class Broker:
         """
         raise NotImplementedError
 
-    def enqueue(self, message, *, delay=None):  # pragma: no cover
+    def enqueue(self, message, *, delay: Optional[int] = None):  # pragma: no cover
         """Enqueue a message on this broker.
 
         Parameters:
@@ -225,7 +225,7 @@ class Broker:
         """
         raise NotImplementedError
 
-    def get_actor(self, actor_name):  # pragma: no cover
+    def get_actor(self, actor_name: str):  # pragma: no cover
         """Look up an actor by its name.
 
         Parameters:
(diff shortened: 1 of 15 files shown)


synweap15:
I don't see anything wrong with `if typing.TYPE_CHECKING`, feel free to go for it. I'd aim for max typing coverage, so I'd suggest to include `message` in the annotation efforts.

I'm waiting for Python 3.9 EOL to get green light for using `type | None` :)

dramatiq/broker.py
```diff
@@ -317,7 +317,7 @@ class Consumer:
     Consumers and their MessageProxies are *not* thread-safe.
     """
 
-    def __iter__(self):  # pragma: no cover
+    def __iter__(self) -> "Consumer":  # pragma: no cover
```
LincolnPuzey, line 320: I think the return type here should be `typing_extensions.Self` because in the subclasses, the return type will be the subclass, not the parent class `Consumer`.

dramatiq/brokers/stub.py
```diff
@@ -157,8 +159,8 @@ def join(self, queue_name, *, fail_fast=False, timeout=None):
         deadline = timeout and time.monotonic() + timeout / 1000
         while True:
             for queue in queues:
-                timeout = deadline and deadline - time.monotonic()
-                join_queue(queue, timeout=timeout)
+                timeout_ = deadline and deadline - time.monotonic()
```
LincolnPuzey, line 162: maybe `timeout_` -> `join_timeout` for clarity?

LincolnPuzey (commented):
Looks good so far, left some initial comments

dramatiq/broker.py
jenstroeger: Unfortunately, [`Self`](https://docs.python.org/3/library/typing.html#typing.Self) was added with 3.11 but this change wants to type-check on 3.9 and up.

LincolnPuzey: I believe if you import it from `typing_extensions` inside of `if TYPE_CHECKING` like [here](https://github.com/Bogdanp/dramatiq/blob/057e03b1727c2813c6e8f5ba209bd3cbec786b0d/dramatiq/actor.py#L41-L42) it should work on 3.9

dramatiq/results/backend.py
```diff
@@ -97,11 +97,11 @@ def get_result(self, message, *, block: bool = False, timeout: typing.Optional[i
             result = self._get(message_key)
             if result is Missing and block:
                 attempts, delay = compute_backoff(attempts, factor=BACKOFF_FACTOR)
-                delay /= 1000
-                if time.monotonic() + delay > end_time:
+                delay_ = delay / 1000
```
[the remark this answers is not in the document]
jenstroeger: Or we’d [allow redefining variables](https://mypy.readthedocs.io/en/stable/config_file.html#confval-allow_redefinition)…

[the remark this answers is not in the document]
synweap15: Here we change the unit so a separate var is justified. Redefining variables is discouraged with an occasional `# mypy: ignore` where appropriate & justified

dramatiq/worker.py
```diff
@@ -332,11 +360,12 @@ def handle_message(self, message):
             message.fail()
             self.post_process_message(message)
 
-    def post_process_message(self, message):
+    def post_process_message(self, message: MessageProxy):
         """Called by worker threads whenever they're done processing
         individual messages, signaling that each message is ready to
         be acked or rejected.
         """
+        assert self.consumer
```
jenstroeger, line 368: Guard this with an `if TYPE_CHECKING` to show that this assertion is for the benefit of static type checking only?

jenstroeger: It still understands `asserts` and uses them for [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html#type-narrowing-expressions). Looks like 
```python
if TYPE_CHECKING:
    assert self.consumer 
```
works just fine and at least makes the intent of the `assert` clearer to the reader. What do you think?

dramatiq/broker.py
```diff
@@ -365,12 +379,24 @@ def close(self) -> None:
 class MessageProxy:
     """Base class for messages returned by :meth:`Broker.consume`."""
 
-    def __init__(self, message):
+    # For the purpose of static type checking we duplicate the fields
+    # and their respective types of the Message here; at runtime the
+    # message's fields are accessed through the __getattr__ method.
+    if TYPE_CHECKING:
```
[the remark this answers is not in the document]
jenstroeger: Fairnuff. I do prefer the guard to make the intent more explicit: the block of type declarations is for static type checking only.

jenstroeger:
I addressed the above comments. Shall we noodle through the middlewares as well, or are they not part of Dramatiq’s external interface? I think I’m easy either way.

dramatiq/worker.py
LincolnPuzey: Ok good point about the type narrowing. This is good then.

LincolnPuzey (approved):
Thanks @jenstroeger Looks good to me.

LincolnPuzey:
>  Shall we noodle through the middlewares as well, or are they not part of Dramatiq’s external interface? I think I’m easy either way.

Lets not add to this PR any more. Another PR for the base middleware class especially would be good though (#521)

PR #734, opened 2025-07-14
Fix off-by-one error in number of rabbitMQ declare/enqueue retries

LincolnPuzey:
Fixes error introduced in #669 
`attempts` variable is counting the number of attempts made. Once this is >= to the max, raise Exception.
Also update log message to make this more obvious

### DIFF
--- a/tests/test_rabbitmq.py
+++ b/tests/test_rabbitmq.py
@@ -12,6 +12,7 @@ from dramatiq.brokers.rabbitmq import (
     RabbitmqBroker,
     URLRabbitmqBroker,
     _IgnoreScaryLogs,
+    MAX_DECLARE_ATTEMPTS,
 )
 from dramatiq.common import current_millis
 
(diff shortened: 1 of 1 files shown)


karolinepauls (approved):

karolinepauls:
This can be tested with:

```
@patch.object(RabbitmqBroker, "_declare_queue", side_effect=pika.exceptions.AMQPConnectionError)
def test_rabbitmq_ensure_queue_retry(mock_declare_queue):
    # Broker doesn't a real RabbitMQ, since we check connection error handling.
    broker = RabbitmqBroker(host="0.0.0.0")

    with pytest.raises(dramatiq.errors.ConnectionClosed):
        broker.declare_queue("some-queue", ensure=True)

    assert mock_declare_queue.call_count == MAX_DECLARE_ATTEMPTS

```

jenstroeger (approved):

LincolnPuzey:
@karolinepauls Good suggestion, I will add a test.

> so the observed behaviour doesn't change.

This is fixing the bug introduced only 6 months ago, so I think changing the behavior back to what it was before that is ok.

PR #735, opened 2025-07-14
Improve typing coverage of middleware classes

### LINKED ISSUE #521: FeatReq: Typehints in dramatiq.middleware

jenstroeger:
Closes #521

Following up on comment https://github.com/Bogdanp/dramatiq/pull/731#issuecomment-3069279831. Alas, I am not familiar enough with Sphinx to work with the `NameError` caused by the missing (typing only) imports (perhaps related issue https://github.com/sphinx-doc/sphinx/issues/13137).

### DIFF
--- a/dramatiq/middleware/middleware.py
+++ b/dramatiq/middleware/middleware.py
@@ -18,6 +18,14 @@
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING, Any, Callable, Optional
+
+if TYPE_CHECKING:
+    from ..actor import Actor
+    from ..broker import Broker, MessageProxy
+    from ..message import Message
+    from ..worker import Worker, _ConsumerThread, _WorkerThread
+
 
 class MiddlewareError(Exception):
     """Base class for middleware errors."""
@@ -36,39 +44,39 @@ class Middleware:
     """
 
     @property
-    def actor_options(self):
+    def actor_options(self) -> set[str]:
         """The set of options that may be configured on each actor."""
         return set()
 
     @property
-    def forks(self):
+    def forks(self) -> list[Callable[[], int]]:
         """A list of functions to run in separate forks of the main
         process.
         """
         return []
 
-    def before_ack(self, broker, message):
+    def before_ack(self, broker: Broker, message: MessageProxy) -> None:
         """Called before a message is acknowledged."""
 
-    def after_ack(self, broker, message):
+    def after_ack(self, broker: Broker, message: MessageProxy) -> None:
         """Called after a message has been acknowledged."""
 
-    def before_nack(self, broker, message):
+    def before_nack(self, broker: Broker, message: MessageProxy) -> None:
         """Called before a message is rejected."""
 
-    def after_nack(self, broker, message):
+    def after_nack(self, broker: Broker, message: MessageProxy) -> None:
         """Called after a message has been rejected."""
 
-    def before_declare_actor(self, broker, actor):
+    def before_declare_actor(self, broker: Broker, actor: Actor) -> None:
         """Called before an actor is declared."""
 
-    def after_declare_actor(self, broker, actor):
+    def after_declare_actor(self, broker: Broker, actor: Actor) -> None:
         """Called after an actor has been declared."""
 
-    def before_declare_queue(self, broker, queue_name):
+    def before_declare_queue(self, broker: Broker, queue_name: str) -> None:
         """Called before a queue is declared."""
 
-    def after_declare_queue(self, broker, queue_name):
+    def after_declare_queue(self, broker: Broker, queue_name: str) -> None:
         """Called after a queue has been declared.
 
         This signals that the queue has been registered with the
@@ -78,19 +86,19 @@ class Middleware:
         them until messages are enqueued or consumed.
         """
 
-    def after_declare_delay_queue(self, broker, queue_name):
+    def after_declare_delay_queue(self, broker: Broker, queue_name: str) -> None:
         """Called after a delay queue has been declared."""
 
-    def before_enqueue(self, broker, message, delay):
+    def before_enqueue(self, broker: Broker, message: Message, delay: int) -> None:
         """Called before a message is enqueued."""
 
-    def after_enqueue(self, broker, message, delay):
+    def after_enqueue(self, broker: Broker, message: Message, delay: int) -> None:
         """Called after a message has been enqueued."""
 
-    def before_delay_message(self, broker, message):
+    def before_delay_message(self, broker: Broker, message: MessageProxy) -> None:
         """Called before a message has been delayed in worker memory."""
 
-    def before_process_message(self, broker, message):
+    def before_process_message(self, broker: Broker, message: MessageProxy) -> None:
         """Called before a message is processed.
 
         Raises:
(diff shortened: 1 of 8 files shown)


dramatiq/middleware/middleware.py
```diff
@@ -33,39 +41,39 @@ class Middleware:
...
 
     @property
-    def actor_options(self):
+    def actor_options(self) -> set[str]:
         """The set of options that may be configured on each actor."""
         return set()
 
     @property
-    def forks(self):
+    def forks(self) -> list[Callable[[], None]]:
```
jenstroeger, line 49: I’ve not dug deeper through the code, but to match Prometheus [here](https://github.com/Bogdanp/dramatiq/blob/057e03b1727c2813c6e8f5ba209bd3cbec786b0d/dramatiq/middleware/prometheus.py#L194-L206) the return type should be `int`.

LincolnPuzey: I think ultimately the fork function is called here

https://github.com/Bogdanp/dramatiq/blob/fb07c5e033f710f5f3e510aaa42ed9bd93737d77/dramatiq/cli.py#L521

So yes I agree the return type should be `int`

dramatiq/middleware/middleware.py
```diff
@@ -36,39 +43,39 @@ class Middleware:
...
-    def after_nack(self, broker, message):
+    def after_nack(self, broker: Broker, message: Message) -> None:
         """Called after a message has been rejected."""
 
-    def before_declare_actor(self, broker, actor):
+    def before_declare_actor(self, broker: Broker, actor: Actor) -> None:
         """Called before an actor is declared."""
 
-    def after_declare_actor(self, broker, actor):
+    def after_declare_actor(self, broker: Broker, actor_name: str) -> None:
```
jenstroeger, line 72: To satisfy `mypy` I had to change the `actor: Actor` argument to `actor_name: str` because https://github.com/Bogdanp/dramatiq/blob/0f193fa666d359aa0baf1972e1f0dafe2bbe679e/dramatiq/broker.py#L183-L184 but this is now inconsistent with `before_declare_actor()`.

I think we should keep this consistent and keep the `actor` arg; thus, we have two options:

1. Change the code in broker to
    ```python
    middleware.after_declare_actor(self, self.actors[actor_name])
    ```
2. Change the `Broker.get_declared_actors()` function to returning Actors instead of actor names:
    ```python
    def get_declared_actors(self) -> set[Actor]:
        return set(self.actors.values())
    ```

Option 2. would break the existing interface but an iota faster; option 1. would work and we’d keep the existing middleware interface. I vote for 1.

LincolnPuzey: Yes I think you should keep the argument as `actor: Actor`, for consistency, and because here

https://github.com/Bogdanp/dramatiq/blob/0f193fa666d359aa0baf1972e1f0dafe2bbe679e/dramatiq/broker.py#L221

The value passed to it is an `Actor`.

I agree with option 1.
Really this is a bug, the `actor` arg was getting a different value (`str`/`Actor`) in the two places this hook is called.
So it would be good to note it as a bugfix in the changelog.

LincolnPuzey:
> Shall we leave the sub-classes alone for now and consider them “internal”, or type them also?

Totally up to you. I would suggest typing the `__init__` methods since those are public interface. (as opposed to the hooks which are called internally by dramatiq).

LincolnPuzey:
(oops clicked wrong button)

dramatiq/middleware/middleware.py
LincolnPuzey: sure that would be great

jenstroeger: PR https://github.com/Bogdanp/dramatiq/pull/743.

jenstroeger:
> I would suggest typing the `__init__` methods since those are public interface. (as opposed to the hook implementations which are called internally by dramatiq).

Agreed, done. This PR requires https://github.com/Bogdanp/dramatiq/pull/743 to pass CI though.

dramatiq/middleware/middleware.py
```diff
@@ -36,39 +44,39 @@ class Middleware:
...
     @property
-    def forks(self):
+    def forks(self) -> list[Callable[[], int]]:
         """A list of functions to run in separate forks of the main
         process.
         """
         return []
 
-    def before_ack(self, broker, message):
+    def before_ack(self, broker: Broker, message: Message) -> None:
```
LincolnPuzey, line 58: I think for before/after ack/nack, it should use `MessageProxy`, rather than `Message`

dramatiq/middleware/middleware.py
```diff
@@ -78,19 +86,19 @@ def after_declare_queue(self, broker, queue_name):
...
-    def before_enqueue(self, broker, message, delay):
+    def before_enqueue(self, broker: Broker, message: Message, delay: int) -> None:
         """Called before a message is enqueued."""
 
-    def after_enqueue(self, broker, message, delay):
+    def after_enqueue(self, broker: Broker, message: Message, delay: int) -> None:
         """Called after a message has been enqueued."""
 
-    def before_delay_message(self, broker, message):
+    def before_delay_message(self, broker: Broker, message: Message) -> None:
```
LincolnPuzey, line 98: I think `MessageProxy` for before_delay_message, before_process_message, after_process_message and after_skip_message

dramatiq/middleware/shutdown.py
```diff
@@ -55,9 +55,11 @@ class ShutdownNotifications(Middleware):
         Defaults to False, meaning actors will not be interrupted, and allowed to finish.
     """
 
-    def __init__(self, notify_shutdown=False):
+    def __init__(self, notify_shutdown: bool = False) -> None:
         self.logger = get_logger(__name__, type(self))
         self.notify_shutdown = notify_shutdown
+
+        self.manager: Union[_GeventShutdownManager, _CtypesShutdownManager]
```
LincolnPuzey, line 62: This line could maybe be simplified to `self.manager: _ShutdownManager`?

dramatiq/middleware/shutdown.py
```diff
@@ -132,34 +134,41 @@ def shutdown(self):
             raise_thread_exception(thread_id, Shutdown)
 
 
-_GeventShutdownManager: Optional[Type[_ShutdownManager]] = None
```
LincolnPuzey, line 135: rather than adding the `if TYPE_CHECKING` block,
can you try changing just this first line to 
`_GeventShutdownManager: Type[_ShutdownManager]`

so rather than `_GeventShutdownManager` being either `None`/the actual class (two different types),
it is undefined / the actual class (which is one type).

dramatiq/middleware/middleware.py
jenstroeger: I have to dig some more through the code but at a first glance I didn’t quite find all the locations where these events are issued by the various brokers… 🤔

dramatiq/middleware/shutdown.py
jenstroeger: Hm. Looking at the code I _assume_ that the intention of
```python
_GeventShutdownManager = None
if is_gevent_active():   
    ...
    _GeventShutdownManager = __GeventShutdownManager
```
is to make the name available for the middleware initialization. But both uses of `_GeventShutdownManager` are guarded by `is_gevent_active()`. So, I removed the module variable `_GeventShutdownManager` and renamed the class `__GeventShutdownManager` to `_GeventShutdownManager` and the code still checks and tests just fine on my local machine:
```
=================================== 175 passed, 131 skipped, 1 warning in 241.83s (0:04:01) ===================================
```
Let me push the changes, and please do take a careful look as well.

LincolnPuzey: hmmm I can't see a reason for `__GeventShutdownManager`.
It is clearly a private name, so I'm happy to get rid of it.

Overall I think this is a good simplification.

LincolnPuzey (approved):
Thanks for addressing all those comments :)

jenstroeger:
@LincolnPuzey do you know when you folks might release the new Dramatiq package with all the typing?

LincolnPuzey:
@jenstroeger There is no planned date. You can always install from the master branch in the meantime

vitosamson:
@LincolnPuzey is there a cadence that you follow for releases or is the process totally ad hoc? We can't install from the master branch for a production application but we'd love to pull in this and other changes. Would it be possible to cut a release sometime in the near future? The last release was three months ago.

PR #736, opened 2025-07-14
Message.mesage_datetime property

karolinepauls:
Using divmod instead of floating point to avoid thinking about precision.

This feature is a trivial convenience that saves on having to check the docs to ensure that `datetime.utcfromtimestamp(message.message_timestamp / 1000)` is correct code.


LincolnPuzey (commented):
Looks reasonable, one comment

LincolnPuzey (approved):

PR #737, opened 2025-07-17
Use `from __future__ import annotations` everywhere

LincolnPuzey:
* For consistency - was already used in some files
* So type hints shouldn't need to use string quotes
* Also make the `tests` directory be checked by isort in CI. (most imports there were already sorted)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,3 +20,5 @@ known_first_party = dramatiq
 multi_line_output = 5
 order_by_type = true
 profile = black
+# Remove annotations import once we only support python 3.14+
+add_imports = from __future__ import annotations
--- a/dramatiq/__init__.py
+++ b/dramatiq/__init__.py
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
 from .actor import Actor, actor
 from .broker import Broker, Consumer, MessageProxy, get_broker, set_broker
 from .composition import group, pipeline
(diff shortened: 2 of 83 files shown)


LincolnPuzey:
Hi @jenstroeger what do you think about this - given you've been doing the typing work lately

setup.cfg
```diff
@@ -20,3 +20,5 @@ known_first_party = dramatiq
 multi_line_output = 5
 order_by_type = true
 profile = black
+# Remove annotations import once we only support python 3.14+
+add_imports = from __future__ import annotations
```
jenstroeger, line 24: Oh this injected the import into every file, huh?

jenstroeger (commented):
> For consistency - was already used in some files

I’m all for consistency, so adding the `annotations` where needed is very agreeable to me. I would probably _not_ import them in files where they’re not used, though (like e.g. [here](https://github.com/Bogdanp/dramatiq/pull/737/files#diff-8ffa1c194fde9a5007e8cafa5a7f55176da8070ccaf7cbcfcd5299478b15a5fc)).

> So type hints shouldn't need to use string quotes

Yup, makes sense.

> Also make the `tests` directory be checked by isort in CI. (most imports there were already sorted)

Separate PR, maybe?

setup.cfg
LincolnPuzey: Yes when you run isort, this means the import will be added automatically.

So rather than needing to think about which files need it or use annotations, its automatically used everywhere, sort of like black formatting :)

jenstroeger: I’m surprised neither flake8 nor pylint notice it as an unused import. But I guess [__future__](https://docs.python.org/3/library/__future__.html) isn’t your standard imported package…

LincolnPuzey: Yes they're really flags to the compiler

LincolnPuzey:
> I would probably not import them in files where they’re not used

I'd rather have isort add them automatically everywhere, than have to think about each file individually.

> Separate PR, maybe?

Could be, but I've done it here so I'll keep it in.

PR #744, opened 2025-07-25
Further improve typing on external interface functions for Broker and Worker

jenstroeger:
Doh, this should’ve been part of PR https://github.com/Bogdanp/dramatiq/pull/731 🤦🏻‍♂️

### DIFF
--- a/dramatiq/broker.py
+++ b/dramatiq/broker.py
@@ -93,7 +93,7 @@ class Broker:
         overwrite when they are declared.
     """
 
-    def __init__(self, middleware: Optional[list[Middleware]] = None):
+    def __init__(self, middleware: Optional[list[Middleware]] = None) -> None:
         self.logger = get_logger(__name__, type(self))
         self.actors: dict[str, Actor] = {}
         self.queues: Any = {}  # Subclasses make this a set!
@@ -132,7 +132,7 @@ class Broker:
         *,
         before: Optional[type[Middleware]] = None,
         after: Optional[type[Middleware]] = None,
-    ):
+    ) -> None:
         """Add a middleware object to this broker.  The middleware is
         appended to the end of the middleware list by default.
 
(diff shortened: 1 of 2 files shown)


dramatiq/broker.py
```diff
@@ -229,7 +229,7 @@ def declare_queue(self, queue_name: str) -> None:  # pragma: no cover
         """
         raise NotImplementedError
 
-    def enqueue(self, message, *, delay: Optional[int] = None):  # pragma: no cover
```
LincolnPuzey, line 232: Can the `message` parameter also be typed?

The `Message` class is generic for the Result type.
And the Message result type won't change when the message gets enqueued,
so strictly speaking I think this method also needs to be generic

```python
def enqueue(self, message: Message[R], *, delay: Optional[int] = None) -> Message[R]
```

LincolnPuzey (commented):
Thanks,

dramatiq/broker.py
jenstroeger: > DO NOT import it. Sorry for yelling, […]

Sheeesus you woke me up from drowsy code doodling!

> […] but importing it is an anti-pattern I have seen before. It gives the illusion that the TypeVar is somehow 'shared' between the different places it is used, when in fact it is not.

Agreed, though I thought I’d ask anyway. With that in mind I always use `_R` or some such sunder-name to denote exactly that “private” purpose.

> side note, if you are supporting 3.12+ you can use the new syntax for generics which makes the question moot. https://docs.python.org/3/whatsnew/3.12.html#pep-695-type-parameter-syntax

Well Dramatiq still supports 3.9+ so our hands are a bit tied.

> Sure, that can be a different PR. I suggest turning on [disallow_any_generics](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_any_generics) and seeing what errors there are. Maybe still add `message: Message` in this PR.

That introduced a lot of noise, let me open an issue in a bit to track that change.

I updated the PR with the `Message` type only.

PR #745, opened 2025-07-28
Replace StrictRedis with Redis and bump redis version to >=4

udogri:


LincolnPuzey:
@udogri Thanks for the PR. I think we need improve our black setup for the project. Let's put this on hold while we do that

LincolnPuzey:
@udogri Can you recreate/rebase this on top of the latest `master` branch? There should be no extra diff from `black` now that #746 is merged.

udogri:
Alright no problem

PR #746, opened 2025-07-31
Improve black setup

LincolnPuzey:
This improves 2 things with our black setup:

1. moves the config to pyproject.toml (the only location black supports).
This means that contributors won't reformat files with a shorter line-length.
2. Makes the CI check the whole repository. This is so various python files aren't missed.


LincolnPuzey:
I'm also in support for migrating `setup.py` to `pyproject.toml`. Any objections @Bogdanp @karolinepauls ?

PR #757, opened 2025-08-11
Add traceback for ConnectionError in ConsumerThread

### LINKED ISSUE #756: Add Traceback Logging for ConnectionError in Worker
In the current implementation, when a ConnectionError is caught in the consumer code, only the [error message](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/worker.py#L304C1-L306C51) is logged without the full traceback.

Without traceback information, it’s challenging to build reliable heuristics to filter these errors in Sentry.

tefra:
Modified the exception handling in worker to include full traceback information for ConnectionError.

Resolves #756


LincolnPuzey:
Actually after testing this, I am not in favor of adding the traceback.

Because it makes the logs from connection errors a lot noisier, and gives the impression of a bug in dramatiq, when connection errors are in fact known/expected errors.
Because starting dramatiq with no broker service running, is a likely scenario for a beginner, we want to avoid this.

PR #758, opened 2025-08-13
Change StubBroker.join() parameter fail_fast to default to True

### LINKED ISSUE #739: Change `fail_fast` default to `True` for the StubBroker `join()` method
As indicated in the docstring here

https://github.com/Bogdanp/dramatiq/blob/1b23e03fc6ba54d05e87938d8a08c92982236338/dramatiq/brokers/stub.py#L132-L150

the `fail_fast` argument should be changed to default to `True`

LincolnPuzey:
When `fail_fast=True` (now the default) and any message gets dead-lettered during the join, then an exception will be raised.

Resolves #739

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -16,6 +16,21 @@ Major Breaking Changes
 
 These are breaking changes we believe are most likely to effect your project.
 
+* The ``fail_fast`` argument to |StubBroker_join| now defaults to True.
+  This means that calling |StubBroker_join|, will by default,
+  re-raise any Exceptions that caused messages to get dead-lettered
+  (i.e. any uncaught Exceptions in your actor functions).
+  You may need to explicitly catch these exception in your tests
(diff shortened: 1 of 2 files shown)


synweap15:
@LincolnPuzey it's a good direction given we're aiming at a major version bump. We'd be consistent, I'd say it's worth it. 

It would be a good reference for developers if the tests caught the exception + at least one test that shows how `fail_fast=False` behaves.

LincolnPuzey:
@synweap15 I'm thinking about projects with a large test suite that might find changing to the new behavior difficult.

How about we add a StubBroker option that controls the `fail_fast` default?
I've added commit a700ba7194b79a5c7d41283c89348b8f55757f41 to show what I mean.

That way large projects can do `StubBroker(fail_fast_default=False)` (presumably in a small number of places (e.g. pytest fixture)), if they want to upgrade to v2, but updating all their tests straight away is too much work.

PR #761, opened 2025-08-19
Fixes #759 Consumers down when ETA is None (or not isnumeric)

### LINKED ISSUE #759: Consumers down when ETA is None (or not isnumeric)
## Checklist

* [x] What OS are you using?
* [x] What version of Python are you using?
* [x] What Broker are you using?
* [x] What did you do?
* [x] What happened?

## What OS are you using?

Ubuntu 20.04.6 LTS (Focal Fossa)

## What version of Python are you using?

3.12.11

1.13

## What Broker are you using?

RabbitMQ

## What did you do?

A colleague sent the following payload manually from the rabbitmq UI in proper JSON format: 
```
{
  "queue_name": "some_queue",
  "actor_name": "some_queue_consumer",
  "args": ["9999"],
  "kwargs": {},
  "options": {
    "retries": 3,
    "eta": null
  },
  "message_id": "1a0227c3-33cf-4e65-ade4-a69aef9db229",
  "message_timestamp": 1754316211252
}
```

When ETA is None the consumer throws a TypeError when comparing the value with current milliseconds:

`if eta > current_millis():`

The consumers should handle this case gracefully and should not die in case the payload is invalid. 

## What happened?

We get the following error: 

TypeError
`'>' not supported between instances of 'NoneType' and 'int'`

The consumer dies and gets rebooted a couple of times until it dies completely and we are left with no consumer processes at all for the specified queue.

## Fix Proposal
I can send a PR but I am not sure on how to handle it. Should we ignore invalid eta values and silently continue or should we raise an exception? I can open a PR once I get some feedback.

gurelkaynak:
Added a check for the eta value and set it to zero in case it's not an int.


karolinepauls:
> I suggested setting eta to 0, but after consideration I'd pick current_milis() instead. What are your thoughts?

Upon a not very deep consideration, I agree that we while shouldn't change the message's queue (normal or delayed), we don't really know which one intended - maybe `eta=None` actually means the message isn't supposed to be delayed? Maybe `eta='1234'` means that it is? However at this point we're reaching a "not worth it" complexity threshold.

Maybe, at most we could map:

- None -> immediate
- integer, float (?) -> delayed
- anything else -> delayed till `current_millis()`, warning log

> From a pro-dev perspective, while handling edge cases like the discussed one keep the state sane, they could ultimately make user code bugs harder to find. Therefore I'd go with keeping the log warning if that's okay.

Besides, what I would like to see in a follow-up PR is an error logged (or even an exception raised) if when sending a message with an incorrect ETA. And making sure we handle negative ETAs.

karolinepauls:
@synweap15 @LincolnPuzey are you OK with the approach I proposed?

Namely this:

> Maybe, at most we could map:
>   - None -> immediate
>   - integer, float (?) -> delayed
>   - anything else -> delayed till current_millis(), warning log

LincolnPuzey:
> are you OK with the approach I proposed?

@karolinepauls Yeah that sounds ok to me. "anything else" case should only effect manually-queued messages that up to now would be errors.

synweap15:
@gurelkaynak thanks!

PR #769, opened 2025-09-06
Print group callback

aequitas:
This PR adds group callbacks to the `group` `__str__` function.

```
>>> import dramatiq
...
... @dramatiq.actor
... def task():
...   pass
...
... group = dramatiq.group([task.message()])
... group.add_completion_callback(task.message())
... print(str(group))
...
group([task()]) -> (task())
```

### DIFF
--- a/dramatiq/composition.py
+++ b/dramatiq/composition.py
@@ -198,7 +198,15 @@ class group:
         return len(self.children)
 
     def __str__(self):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


LincolnPuzey:
OK thanks, but I'd rather not make this change. To me the small benefit is not worth the (also small) risk of this being a disruptive change to someone.

PR #771, opened 2025-09-10
redis: Remove worker heartbeat compatibility code

### LINKED ISSUE #770: Remove worker heartbeat compatibility code
Essentially this is about reverting https://github.com/Bogdanp/dramatiq/commit/d6cff30701c7338ce1e4ef3546ba2fab9fe12bd9. The deprecation notices in that one had no expiration date, but it seems to me that 7+ years should have been enough for anyone interested to migrate :) Especially since it looks like v2.0.0 is in the works.

mikeroll:
This partially reverts commit d6cff30701c7338ce1e4ef3546ba2fab9fe12bd9.

The heartbeat logic had been changed in [v1.2.0](https://github.com/Bogdanp/dramatiq/releases/tag/v1.2.0), so removing this is a breaking change only for those who would be upgrading from [v1.1.0](https://github.com/Bogdanp/dramatiq/releases/tag/v1.1.0) - which released more than 7 years ago - directly to v2.

Closes #770

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -50,7 +50,11 @@ These are changes that while technically breaking, we believe are unlikely to ef
 * The deprecated ``requeue_deadline`` and ``requeue_interval`` arguments of |RedisBroker| have been removed.
   These have been deprecated and have had no effect since version 1.2.0.
   (`#782`_, `@mikeroll`_)
-* |RedisBroker|: the obsolete worker heartbeat compatibility code has been removed.
+* |RedisBroker|: The code for compatibility with pre-v1.2.0 acks has been removed.
+  If you are using the |RedisBroker|, you must first upgrade Dramatiq to a version >=1.2.0, <2.0.0,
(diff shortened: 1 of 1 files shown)


docs/source/changelog.rst
```diff
@@ -19,12 +19,16 @@ Breaking Changes
 * The ``backend`` argument to the |Results| middleware is now required.
   Previously, not supplying this argument would result in a non-functional |Results| middleware.
   (`#728`_, `@LincolnPuzey`_)
+* The long-deprecated ``requeue_{deadline,interval}`` parameters of |RedisBroker|,
+  as well as the obsolete heartbeat compat code, have been removed. (`#771`_, `@mikeroll`_)
```
mikeroll, line 23: Not _that_ breaking I suspect but technically we're still changing the interface.

LincolnPuzey:
@mikeroll Can you split removing the Broker args into its own PR? I am happy to merge that. Removing the other compat code might need more discussion.

PR #773, opened 2025-09-12
add dramatiq_worker_timeout environment variable

ksoviero-zengrc:
This PR allows specifying the `dramatiq_worker_timeout` as an environment variable which is then passed into the worker. The goal for us was to be able to increase the `max_backoff` setting in `compute_backoff` with Redis because the default of 1000ms is too low for our use case.


LincolnPuzey (changes requested):
Thanks, I want to avoid new command line options for rarely-used configuration. Can you make this configurable from a env var, similar to `dramatiq_restart_delay` and `dramatiq_queue_prefetch`

PR #774, opened 2025-09-15
Document dynamic time limits for Dramatiq actors

albcunha:
Added examples for setting dynamic time limits in Dramatiq, as it is a common situation in OCR and time varying pipelines.

### DIFF
--- a/docs/source/guide.rst
+++ b/docs/source/guide.rst
@@ -319,6 +319,56 @@ wrap its source code in a try block and catch |TimeLimitExceeded|::
       except TimeLimitExceeded:
           teardown_missiles()  # <- this will run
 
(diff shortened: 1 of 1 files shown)


LincolnPuzey (changes requested):
@albcunha Thanks, however this needs some changes before we would accept it.

1. all the new docs should be moved to `cookbook.rst`. `guide.rst` is only for basic/essential topics relevant to first-time setup.
2. The example needs to be made more generic. Do not reference "pages" or other details of your application.
3. Put it all under a docs heading of "Custom Middleware" and approach it from a direction of "here is an example of a custom middleware".

albcunha:
Ok! I'm glad you liked it. I will make the changes

PR #775, opened 2025-09-17
Deprecate the `dramatiq_group_callback_barrier_ttl` environment variable

mikeroll:
I don't think this needs to exist. `GroupCallbacks` is not in the default middlewares and has to be added and configured explicitly anyway, might as well override `barrier_ttl` too if desired.

Also added `GroupCallbacks` to the reference while I was at it.


mikeroll:
Not really a specific problem, I just find the current set of env vars somewhat confusing (as in, _why exactly_ is this an env var and not a class/function/cli argument?). This is one of the more obvious cases IMO.

I would add a deprecation warning if you agree.

PR #784, opened 2025-10-07
Drop python 3.9 support

LincolnPuzey:


synweap15:
Personally, I'd keep it around for a bit more, the reality is many projects don't keep up with EOLs. However, this is the proper action formally.

Let's drop it.

PR #785, opened 2025-10-07
Run tests on python 3.14 with free threads

jairhenrique:


LincolnPuzey:
Yes I believe that using gevent on free-threaded python is a no-go (https://www.gevent.org/install.html#supported-platforms).

Perhaps right at the start of the `dramatiq-gevent` entrypoint we can add a check and raise an exception if so.

@synweap15 Where did you get `is_nogil_mode` attribute from? I can't see it mentioned in the docs.

synweap15:
@LincolnPuzey sorry, I recently work with a custom-patched CPython where `is_nogil_mode` is exposed.

The offical CPython way has been introduced in  https://github.com/python/cpython/pull/118514 - `sys._is_gil_enabled()`, reflecting the current state. I have updated my comment for clarity.

LincolnPuzey:
@spumer Thanks for the suggestion, I hadn't seen those plugins.

LincolnPuzey:
Thanks @jairhenrique, but I'm closing this in favor of #813

Issue #791, opened 2025-10-29
Raising TimeoutError in a async task causes infinite loop and process death


LucasAdamsTR:
## What OS are you using?

Mac, debian

## What version of Python are you using?

3.12.8

## What version of Dramatiq are you using?

1.17.1

## What Broker are you using?

Redis

## What did you do?

One of our tasks can raise a TimeoutError for a specific reason. This error is not caught or handled by our task code, so it propagates to dramatiq. When this was raised, we found that the thread would stop processing tasks, and soon (1 minute) the entire container would die with 100% CPU and OOM. I reproduced this on my mac laptop with a simple example:
```
@dramatiq.actor
async def test():
    raise TimeoutError
````

## What happened?
The issue can be traced to how dramatiq [polls for future completion](https://github.com/Bogdanp/dramatiq/blob/73969e8146dbb97a8203c373fd13002a806e1d3f/dramatiq/asyncio.py#L149) on the event loop: 
```
            while True:
                try:
                    # Use a timeout to be able to catch asynchronously
                    # raised dramatiq exceptions (Interrupt).
                    return future.result(timeout=self.interrupt_check_ival)
                except concurrent.futures.TimeoutError:
                    continue
```
The `future.result` call can throw a TimeoutError in 2 scenarios:
- the task is still running, and we have slept for `interrupt_check_ival`
- the task is no longer running, because it threw TimeoutError itself
In the second case, this block of code turns into an infinite loop with *no sleep*. future.result will always immediately return with TimeoutError after this. It will run at 100% CPU and brick your process. 

This could be fixed by checking whether the future is done or not before continuing:
```
try:
    return future.result(timeout=self.interrupt_check_ival)
except concurrent.futures.TimeoutError:
    if future.done():  # then this was the task's own TimeoutError
        # re-raise the underlying exception
        return future.result()   # will raise the original exception
    continue
```

LincolnPuzey:
Thanks for the report! Looks like a real problem to me

LincolnPuzey:
@synweap15 I Have opened #801 with a modified version of the fix. No tests yet

PR #806, opened 2025-12-17
AsyncIO: avoid holding exceptions and traceback

frankie567:
More complete and reasonable fix for the problem I described in #802.

## Problem description

In situations with lot of tasks being executed, and some of them raising exceptions with large tracebacks, we might encounter huge spikes in memory consumption that might lead to OOM issues. Mainly, this is because the garbage collector may not run timely enough under high pressure. In a sense, this is similar to the old issue #351.

The problem has several sources:

1. The coroutine retains its exception context until it's garbage collected:

https://github.com/Bogdanp/dramatiq/blob/aa91cdfcfa6d8ad957ca0afe900266617f2661f8/dramatiq/asyncio.py#L137-L141

2. The future retains the exception context and traceback raised by the coroutine until it's garbage collected:

https://github.com/Bogdanp/dramatiq/blob/aa91cdfcfa6d8ad957ca0afe900266617f2661f8/dramatiq/asyncio.py#L143-L149

## Proposed solution

The solution is to make sure to detach the exception context from the coroutine and future so we can manually manage its lifecycle and force garbage collection. Basically, we use local variables to store the result and exception of the coroutine; instead of relying on `future.result()` directly.

I also added a unit test for this. Admittedly, it's a bit complex since it's hard to replicate this behaviour inside pytest with the stub broker.

## Side note

The happy side-effect of this change is that it makes the TimeoutError fix (#791) a bit more readable, since we directly handle the `TimeoutError` raised inside the coroutine.

### DIFF
--- a/tests/middleware/test_asyncio.py
+++ b/tests/middleware/test_asyncio.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import asyncio
+import logging
+import weakref
 from threading import get_ident
 from unittest import mock
 
@@ -13,6 +15,7 @@ from dramatiq.asyncio import (
     get_event_loop_thread,
     set_event_loop_thread,
 )
+from dramatiq.brokers.stub import StubBroker
 from dramatiq.logging import get_logger
 from dramatiq.middleware import CurrentMessage
 from dramatiq.middleware.asyncio import AsyncIO
@@ -82,7 +85,9 @@ def test_event_loop_thread_run_coroutine_exception(started_thread: EventLoopThre
     assert e.traceback[-1].name == "raise_actual_error"
 
 
-def test_event_loop_thread_run_coroutine_timeout_exception(started_thread: EventLoopThread):
+def test_event_loop_thread_run_coroutine_timeout_exception(
+    started_thread: EventLoopThread,
+):
     """Test that TimeoutError in coroutine doesn't lead to infinite loop.
 
     Regression test for https://github.com/Bogdanp/dramatiq/issues/791
@@ -98,6 +103,49 @@ def test_event_loop_thread_run_coroutine_timeout_exception(started_thread: Event
 
     with pytest.raises(TimeoutError, match="something took too long"):
         started_thread.run_coroutine(coro)
+
+
+def test_run_coroutine_exception_doesnt_leak(stub_broker: StubBroker, caplog: pytest.LogCaptureFixture):
+    # Disable log capturing to prevent pytest from holding traceback references
+    # via captured LogRecord.exc_info
+    caplog.set_level(logging.CRITICAL)
+
+    stub_broker.add_middleware(AsyncIO())
+
+    class Payload:
+        __slots__ = ("data",)
+
+        def __init__(self, data: bytes) -> None:
+            self.data = data
+
+    class PayloadError(Exception):
+        def __init__(self, payload: Payload) -> None:
+            self.payload = payload
+            super().__init__(payload)
+
+    weak_refs: list[weakref.ref[Payload]] = []
+
+    @actor(max_retries=1, max_backoff=1)
+    async def failing_actor():
+        payload = Payload(b"x" * 1024)
+        weak_refs.append(weakref.ref(payload))
+        raise PayloadError(payload)
+
+    with worker(stub_broker, worker_timeout=100, worker_threads=1) as stub_worker:
+        failing_actor.send()
+        stub_broker.join(failing_actor.queue_name, fail_fast=False)
+        stub_worker.join()
+
+    # StubBroker overloads MessageProxy.clear_exception to do nothing.
+    # Simulate here the normal behavior of clearing exception references.
+    for message in stub_broker.dead_letters:
+        del message._exception
+
+    stub_broker.flush_all()
+
+    # Check that payloads were collected
+    for weak_ref in weak_refs:
+        assert weak_ref() is None, "Payload object still alive"
 
 
 @pytest.mark.skipif(
--- a/dramatiq/asyncio.py
+++ b/dramatiq/asyncio.py
@@ -20,8 +20,10 @@ from __future__ import annotations
 import asyncio
 import concurrent.futures
 import functools
+import gc
 import logging
 import threading
+from types import TracebackType
 from typing import Awaitable, Callable, Optional, ParamSpec, TypeVar
 
 from .threading import Interrupt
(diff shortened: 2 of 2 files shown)


frankie567:
I'm not sure. I've ran my test script in two scenarios: high pressure (one task every 100 ms) and low pressure (one task every 1 second). Here are the results:

**High pressure**

Figure_fast

**Low pressure**

Figure_slow

In both cases, we see the memory grow rapidly but stabilises at some point (but at a very high value). In the low pressure scenario, we do see some drops which shows that we free up memory at some point, but we're still on that "plateau". My reading is that the EventLoopThread might retain the objects from the last run until a new run overrides them.

---

With the fix in this PR, here is the memory consumption with the same script (one task every 100 ms):

Figure_fast_fix

frankie567:
Sure, here it is:

```py
import os
import time

import dramatiq
import psutil
from dramatiq.brokers.redis import RedisBroker
from dramatiq.middleware.asyncio import AsyncIO

MAX_BACKOFF = 100 # Adjust the pressure with this constant
MEMORY_LOG_FILE = "memory_usage.csv"

broker = RedisBroker(url="redis://localhost:6379/0")
dramatiq.set_broker(broker)
broker.add_middleware(AsyncIO())


class BigException(Exception):
    def __init__(self, a: bytes) -> None:
        self.a = a
        super().__init__("Big exception")


def log_memory(label: str = "") -> None:
    process = psutil.Process()
    memory_info = process.memory_info()
    timestamp = time.time()
    memory_mb = memory_info.rss / 1024 / 1024

    file_exists = os.path.exists(MEMORY_LOG_FILE)
    with open(MEMORY_LOG_FILE, "a") as f:
        if not file_exists:
            f.write("timestamp,memory_mb,label\n")
        f.write(f"{timestamp},{memory_mb:.2f},{label}\n")


@dramatiq.actor(actor_name="oom_task", max_retries=1_000_000, max_backoff= MAX_BACKOFF)
async def oom_task() -> None:
    log_memory("before_alloc")
    a = bytes(bytearray(128 * 1024 * 1024))
    log_memory("after_alloc")
    raise BigException(a)


if __name__ == "__main__":
    oom_task.send()
```

One-liner to plot the graph:

```sh
uv run --with pandas --with matplotlib python -c "import pandas as pd; import matplotlib.pyplot as plt; df = pd.read_csv('memory_usage.csv'); plt.plot(df['timestamp'] - df['timestamp'].iloc[0], df['memory_mb'], marker='o'); plt.xlabel('Time (s)'); plt.ylabel('Memory (MB)'); plt.title('Memory Usage'); plt.show()"
```

themavik (commented):
Routing failures through exception_container and re-raising with cleared traceback refs matches the weakref regression goal. nit: import gc is unused in the diff—drop it unless a follow-up commit uses it.

LincolnPuzey:
@frankie567 I've done some testing with your examples. I was able to reproduce but my graphs look a bit different to yours:

e.g. High pressure running with python 3.14.6

Figure_1_314

Note python 3.14.0 - 3.14.4 used a different default GC, which was reverted in 3.14.5: https://docs.python.org/3.14/whatsnew/3.14.html#notable-changes-in-3-14-5
If you were using a version in that range could explain the difference?

Anyway from my graph, it looks like there are reference cycles being created that include the Exception object, so they are only freed when the generational gargage collector runs (about every 15 seconds). This is the big drops.

I think the reference cycle is:

```text
BigException -> trackback -> frames -> EventLoopThread.run_coroutine locals -> Future -> BigException
```

Looking at the printout of the trackback:

```text
  File "/home/lincoln/dramatiq/dramatiq/worker.py", line 560, in process_message
    res = actor(*message.args, **message.kwargs)
  File "/home/lincoln/dramatiq/dramatiq/actor.py", line 194, in __call__
    return self.fn(*args, **kwargs)
           ~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/lincoln/dramatiq/dramatiq/asyncio.py", line 69, in wrapper
    return event_loop_thread.run_coroutine(async_fn(*args, **kwargs))
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lincoln/dramatiq/dramatiq/asyncio.py", line 149, in run_coroutine
    return future.result(timeout=self.interrupt_check_ival)
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lincoln/.local/share/uv/python/cpython-3.14.6-linux-x86_64-gnu/lib/python3.14/concurrent/futures/_base.py", line 454, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/home/lincoln/.local/share/uv/python/cpython-3.14.6-linux-x86_64-gnu/lib/python3.14/concurrent/futures/_base.py", line 396, in __get_result
    raise self._exception
  File "/home/lincoln/dramatiq/dramatiq/asyncio.py", line 139, in wrapped_coro
    return await coro
           ^^^^^^^^^^
  File "/home/lincoln/dramatiq/app.py", line 75, in oom_task
    raise BigException(a)
app.BigException: Big exception
```

I think the `line 149, in run_coroutine` frame is the one who's locals reference the Future (`future` variable), which then holds a reference back to the Exception.

tests/middleware/test_asyncio.py
```diff
@@ -100,6 +105,49 @@ async def raise_error():
         started_thread.run_coroutine(coro)
 
 
+def test_run_coroutine_exception_doesnt_leak(stub_broker: StubBroker, caplog: pytest.LogCaptureFixture):
```
LincolnPuzey, line 108: Running this new test against `master` branch, it passes. I would expect it to fail if it was properly testing the issue.

Of course writing an effective regression test for this sort of issue is pretty tricky.

frankie567: Pretty sure it was failing before my fix 😄 But as you mentioned, the GC change in recent version may have corrected the problem indirectly.

tests/middleware/test_asyncio.py
```diff
@@ -100,6 +105,49 @@ async def raise_error():
...
+    # Disable log capturing to prevent pytest from holding traceback references
+    # via captured LogRecord.exc_info
+    caplog.set_level(logging.CRITICAL)
+
+    stub_broker.add_middleware(AsyncIO())
+
+    class Payload:
+        __slots__ = ("data",)
```
LincolnPuzey, line 116: Using `__slots__` here is preventing creating weakrefs, so the `weak_refs` list is always empty.

LincolnPuzey:
@frankie567 No need to fix my comments, they're really just notes for me. I'm working on an alternate fix.

LincolnPuzey:
Alternate fix merged in #878.

Thanks again for the scripts to reproduce and measure.

Issue #811, opened 2026-01-01
Issue warning when gevent used with free-threaded python


LincolnPuzey:
> Reading the gevent docs, though, my understanding is that using gevent will simply enable the GIL, so maybe a warning would be more appropriate when `dramatiq-gevent` is run in a free-threading Python.

_Originally posted by @Bogdanp in https://github.com/Bogdanp/dramatiq/issues/785#issuecomment-3382483444_
            
<hr/>

We should add a check to detect free-threaded python in the gevent entry point files (`dramatiq-gevent` and `pytest-gevent.py`), and issue a python warning when it is detected.

PR #815, opened 2026-01-09
chore: show warning when gevent is used with free-threaded Python

synweap15:
Closes https://github.com/Bogdanp/dramatiq/issues/811

### DIFF
--- a/bin/dramatiq-gevent
+++ b/bin/dramatiq-gevent
@@ -2,13 +2,18 @@
 import sys
 import warnings
 
-# Check for free-threaded Python before gevent patching
-if hasattr(sys, "_is_gil_enabled") and not sys._is_gil_enabled():  # noqa: SLF001
+# Check for a free-threaded Python build (PEP 703) before gevent patching.
+#
+# Note: a free-threaded build may already have the GIL enabled at this point
+# (e.g. via early imports), so we base this on the build rather than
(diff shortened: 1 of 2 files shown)


bin/dramatiq-gevent
```diff
@@ -1,13 +1,24 @@
 #!python
+import sys
+import warnings
+
+# Check for free-threaded Python before gevent patching
+if hasattr(sys, "_is_gil_enabled") and not sys._is_gil_enabled():  # noqa: SLF001
```
LincolnPuzey: perhaps we should be checking if the build is free-threaded, not if the GIL is disabled. These are technically different things. The GIL may already be disabled somehow.

`if "free-threading" in sys.version: ...`

https://docs.python.org/3/howto/free-threading-python.html#identifying-free-threaded-python

pytest-gevent.py
```diff
@@ -1,16 +1,25 @@
...
 try:
     from gevent import monkey
 
-    monkey.patch_all()  # noqa
+    monkey.patch_all()
 except ImportError:
-    import sys
-
-    sys.stderr.write("error: gevent is missing. Run `pip install gevent`.")
+    sys.stderr.write("error: gevent is missing. Run `pip install gevent`")
```
LincolnPuzey: no need to change this line?

Issue #827, opened 2026-01-27
Support/Mitigate RabbitMQ `consumer_timeout` behavior


davidt99:
In RabbitMQ 3.8.15+, a new concept was introduced: consumer timeout. If a message is not acknowledged within the timeout period (defaulting to 30 minutes), RabbitMQ closes the connection and returns the message to the queue. This timeout can be configured globally, per-queue, or disabled (though disabling it is not recommended).

This behavior affects Dramatiq in several ways:

## 1. Reprocessing loops
If the `TimeLimit` middleware is not configured, or if its limit is higher than the RabbitMQ consumer timeout, a message that hits the timeout will be reprocessed indefinitely because Dramatiq loses the ability to ACK it once the connection drops.
A bug causing a worker to hang will eventually propagate to all workers as they all pick up the "poison" message, attempt to process it, and time out.

We should improve documentation regarding this. Additionally, I suggest adding a warning log when booting a RabbitMQ broker without the `TimeLimit` middleware. Adding `TimeLimit` by default seems incorrect as we cannot programmatically fetch the server's timeout value.

## 2. Delays exceeding the timeout
If a message has a delay (or backoff) longer than the consumer timeout, the Delayed Delivery (DQ) consumer thread will trigger a timeout. This causes all messages currently held in that thread to be returned to the queue. While this doesn't stop task processing entirely, it creates an unhealthy "churn" cycle.

I see here two changes we probably need to make:
1. Update `DEFAULT_MAX_BACKOFF` to 30 minutes when using the RabbitMQ broker to align with default server settings.
2. Modify the DQ consumer thread to re-enqueue messages at specific intervals (e.g., every 15 minutes) to avoid hitting the timeout. This is also a prerequisite for supporting Quorum Queues properly in the future.

## 3. Visibility and Logging
As discussed in #825, current logging can be misleading. In my experience, hunting down the specific message that exceeded 30 minutes was difficult because the exception is raised (and the connection closed) before the task actually finishes. We should improve the clarity of the logs to help users identify when a RabbitMQ-enforced timeout has occurred.

davidt99:
[...]

PR #869, opened 2026-06-18
Add QuorumRabbitmqBroker (RabbitMQ 4.3+)

davidt99:
## Summary

Adds `QuorumRabbitmqBroker`, a drop-in `RabbitmqBroker` subclass that declares its queues as [quorum queues](https://www.rabbitmq.com/docs/quorum-queues). It reuses the existing delay/retry/worker machinery and only adapts to the three quorum-queue features below. Requires **RabbitMQ 4.3+**.

### Consumer timeout

A delivery left unacked past the broker's `consumer_timeout` (default 30 min) gets requeued. dramatiq holds delayed messages unacked in memory until their eta, so long delays would trip it. The new optional `consumer_timeout` (default `None` = unchanged on the classic broker; 30 min on quorum) is sent as an `x-consumer-timeout` consumer argument on the **delay-queue consumer only**; the worker *re-leases* a held message at `0.75×` that value — re-enqueuing it with its remaining delay so the eta is preserved and it fires once.

### Poison-message handling

Quorum queues dead-letter after `x-delivery-limit` redeliveries. We keep RabbitMQ's default on the main queue (overridable via `delivery_limit`) but disable it on the delay queue (`-1`), whose messages are intentionally held and re-leased.

### Priority — and why 4.3

Quorum queues reject the classic `x-max-priority` and provide built-in priority: only 2 levels in 4.0–4.2, full **0–31 strict priorities in 4.3**. We require 4.3 so `broker_priority` keeps equivalent ordering; `max_priority` is rejected and unset priority defaults to `4` (vs `0` on classic).

### Native delayed-retry (not used)

4.3 also offers native delayed retries, but the delay is queue-level and per-message overrides need AMQP 1.0 (unsupported by pika), so dramatiq's app-side retries/delay queue are kept unchanged.

---

> ⚠️ **Draft — not yet tested in a real environment.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/source/global.rst
+++ b/docs/source/global.rst
@@ -21,6 +21,7 @@
 .. |Prometheus| replace:: :class:`Prometheus<dramatiq.middleware.prometheus.Prometheus>`
 .. |RabbitmqBroker_join| replace:: :meth:`join<dramatiq.brokers.rabbitmq.RabbitmqBroker.join>`
 .. |RabbitmqBroker| replace:: :class:`RabbitmqBroker<dramatiq.brokers.rabbitmq.RabbitmqBroker>`
+.. |QuorumRabbitmqBroker| replace:: :class:`QuorumRabbitmqBroker<dramatiq.brokers.rabbitmq.QuorumRabbitmqBroker>`
 .. |RateLimitExceeded| replace:: :class:`RateLimitExceeded<dramatiq.RateLimitExceeded>`
 .. |RateLimiters| replace:: :class:`RateLimiters<dramatiq.rate_limits.RateLimiter>`
 .. |RedisBroker| replace:: :class:`RedisBroker<dramatiq.brokers.redis.RedisBroker>`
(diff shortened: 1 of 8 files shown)


davidt99:
Will fix #827 as well

davidt99:
@LincolnPuzey would appreciate your feedback on the implementation before I move to testing.

LincolnPuzey:
Hi @davidt99, we aren't planning on adding any new Broker classes to the dramatiq core. You're welcome to publish this as a third-party package yourself.

Thanks, Lincoln

davidt99:
@LincolnPuzey is quorum-queues is something we want to support? I can move the logic inside the current RabbitMQ's broker and add a feature flag to distinguish the between the two modes.
I also fixed the consumer time that exists in all of rabbitmq's version from 3.8.15, regardless of the queue type, which it sits on the worker side.

PR #871, opened 2026-06-24
Add quorum queue support to RabbitmqBroker (queue_type flag) + delay-queue re-lease

davidt99:
## Summary

Follow-up to #869. Per the feedback there (no new `Broker` classes in core),
this folds quorum-queue support into the existing `RabbitmqBroker` behind a
`queue_type` flag instead of a subclass, and resolves the delay-queue half of
#827. `queue_type="classic"` stays the default — today's behavior is unchanged.

### Delay-queue re-lease — `consumer_timeout` (#827)

A delivery left unacked past RabbitMQ's `consumer_timeout` (default 30 min,
enforced since 3.8.15 on any queue type) is requeued, which churns the delay
queue for delays longer than the timeout. The worker now *re-leases* a held
message at `0.75×` `consumer_timeout`, re-enqueuing it with its remaining delay
so the eta is preserved and it fires once. Worker-side; works on classic and
quorum. Defaults to `None` (unchanged) on classic, 30 min on quorum.

### `queue_type="quorum"` (RabbitMQ 4.3+)

Declares the queues as [quorum queues](https://www.rabbitmq.com/docs/quorum-queues),
reusing all the existing delay/retry machinery. Adds poison-message handling
(`x-delivery-limit`, overridable via `delivery_limit`; disabled on the delay
queue) and built-in `broker_priority` (0–31 strict, hence the 4.3 requirement;
`max_priority` is rejected).

---

Addresses #827.

> ⚠️ Not yet tested in a real (production) environment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/source/guide.rst
+++ b/docs/source/guide.rst
@@ -415,6 +415,58 @@ execution::
   rabbitmq_broker = RabbitmqBroker(host="rabbitmq")
   dramatiq.set_broker(rabbitmq_broker)
 
+Quorum Queue Broker
+^^^^^^^^^^^^^^^^^^^
+
+Pass ``queue_type="quorum"`` to the |RabbitmqBroker| to declare its
+queues as `quorum queues`_ instead of classic queues.  Quorum queues are
+replicated, always-durable FIFO queues and are the recommended queue type
+for most workloads.  This requires **RabbitMQ 4.3 or later**::
+
+  import dramatiq
+
+  from dramatiq.brokers.rabbitmq import RabbitmqBroker
+
+
+  rabbitmq_broker = RabbitmqBroker(host="rabbitmq", queue_type="quorum")
+  dramatiq.set_broker(rabbitmq_broker)
+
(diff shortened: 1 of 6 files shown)


davidt99:
@LincolnPuzey I changed the implementation as requested.
If quorum queue isn't something you want to support, please let me know so I won't invest time trying to contribute.

LincolnPuzey:
Hi @davidt99, sorry, yes what I really meant was that we aren't planning to support quorum queues in dramatiq core.

davidt99:
@LincolnPuzey Can I add the requeue mechanism at least? That's affect all rabbit versions from 3.8.15?

LincolnPuzey:
Is that a fix for #827 ?

davidt99:
Yes

LincolnPuzey:
Sure, open a standalone PR for that. Keeping PRs limited in scope helps us review :)

PR #876, opened 2026-07-09
Fix Prometheus delayed_messages_inprogress gauge never returning to zero (#672)

### LINKED ISSUE #672: Prometheus middleware dramatiq_delayed_messages_inprogress metric does not work
# Issues

The prometheus middleware exports a gauge metric `dramatiq_delayed_messages_inprogress` that is documented to track "The number of delayed messages in memory.". I noticed something appeared to be wrong with this metric because in my system the metric never goes down.

My setup:

dramatiq: 1.15.0
broker: redis
workers: 4
processes per worker: 2
threads per process: 1

Looking over the code I can see that the middleware is listening to the `before_delay_message` and `before_process_message` to track when a message is first delayed and then complete by tracking the message IDs. Looking at the actual worker implementation, however, it looks like messages get re-queued back in the broker when their eta arrives instead of being processed directly on the same worker process. (I also wonder if the message retains its same message ID upon being requeued, unsure)

All this seems to create a situation where the process that first marked a message as in progress to be unlikely to be the process that observes its completion, resulting in the metric not being tracked correctly. Additionally this whole process appears to leak memory (albeit slowly and not much) as the delayed message ID set grows over time.

Would be happy to contribute/discuss a patch to fix this metric if desired!

apoorvdarshan:
Closes #672

### Root cause

The Prometheus middleware tracks `dramatiq_delayed_messages_inprogress` by incrementing the gauge in `before_delay_message` and decrementing it in `before_process_message`, keyed on `(queue_name, actor_name)` labels.

A delayed message lives on the `".DQ"` queue while it waits for its eta, so when `before_delay_message` fires the message's `queue_name` is e.g. `default.DQ`, and the gauge is incremented for the label `("default.DQ", actor)`.

When the eta passes, `Worker.handle_delayed_messages` re-enqueues the message onto the **canonical** queue using `q_name()` (which strips the `.DQ` suffix). The re-enqueued message keeps its `message_id`, so `before_process_message` matches it and decrements the gauge — but now the message's `queue_name` is `default`, so the decrement lands on the label `("default", actor)`.

Because the increment and decrement use **different `queue_name` labels**, the `".DQ"` series is only ever incremented and never settles back to zero (the "metric never goes down" behavior reported in the issue), while the canonical series drifts negative.

### Reproduction

Driving a real `Worker` with a `StubBroker` and the Prometheus middleware, sending one delayed message and letting it run to completion, then dumping every child of the gauge:

```python
do_work.send_with_options(delay=100)
broker.join(do_work.queue_name, timeout=5000)
worker.join()

for labels, child in prometheus.inprogress_delayed_messages._metrics.items():
    print(labels, "=", child._value.get())
```

Observed on `master` (before the fix) — the message is delayed on `default.DQ` but processed on `default`, so the two events hit different labels:

```
before_delay   queue=default.DQ  actor=do_work
before_process queue=default     actor=do_work

('default.DQ', 'do_work') = 1.0     <-- only ever incremented, never returns to zero
('default', 'do_work')    = -1.0
```

After the fix (both events recorded against the canonical queue name):

```
before_delay   queue=default.DQ  actor=do_work
before_process queue=default     actor=do_work

('default', 'do_work') = 0.0        <-- gauge settles back to zero
```

### Fix

In `before_delay_message`, normalize the queue name with the existing `q_name()` helper before building the label tuple — the same helper the worker already uses when re-enqueueing the delayed message. This makes the increment and decrement land on the same label set. The change is a one-line label normalization plus the import; it does not alter the id-tracking or any other metric.

### Tests

Added `test_prometheus_middleware_delayed_messages_inprogress_returns_to_zero` in `tests/middleware/test_prometheus.py`. It runs a delayed message through a real worker and asserts every `inprogress_delayed_messages` child returns to `0` (and that the in-memory `delayed_messages` set does not leak). The test **fails on `master`** (the `('default.DQ', 'do_work')` child is stuck at `1.0`) and passes with this change.

The middleware test module and a broad stub/worker/message/common smoke subset pass locally, and `flake8`, `black`, `isort` and `mypy` are clean on the modified files.

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -80,3 +80,4 @@ of those changes to CLEARTYPE SRL.
 | [@ABolouk](https://github.com/ABolouk)                 | Amirhossein Bolouk Asli|
 | [@JoaoAlmeida20](https://github.com/JoaoAlmeida20)     | João Almeida           |
 | [@albcunha](https://github.com/albcunha)               | Alberto Cunha          |
+| [@apoorvdarshan](https://github.com/apoorvdarshan)     | Apoorv Darshan         |
--- a/dramatiq/middleware/prometheus.py
+++ b/dramatiq/middleware/prometheus.py
@@ -21,7 +21,7 @@ import os
 import tempfile
 from http.server import BaseHTTPRequestHandler, HTTPServer
 
-from ..common import current_millis
+from ..common import current_millis, q_name
 from ..logging import get_logger
 from .middleware import Middleware
 
@@ -145,7 +145,13 @@ class Prometheus(Middleware):
             self.total_retried_messages.labels(*labels).inc()
 
     def before_delay_message(self, broker, message):
-        labels = (message.queue_name, message.actor_name)
+        # Delayed messages are held on the ".DQ" queue, but they are
+        # re-enqueued onto (and processed from) the canonical queue, where
+        # before_process_message decrements this gauge.  Use the canonical
+        # queue name here so the inc and dec land on the same label set;
+        # otherwise the ".DQ" series only ever increments and never settles
+        # back to zero.
+        labels = (q_name(message.queue_name), message.actor_name)
         self.delayed_messages.add(message.message_id)
         self.inprogress_delayed_messages.labels(*labels).inc()
 
--- a/tests/middleware/test_prometheus.py
+++ b/tests/middleware/test_prometheus.py
@@ -4,11 +4,16 @@ import time
 import urllib.request as request
 from threading import Thread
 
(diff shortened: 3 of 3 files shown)


LincolnPuzey:
You claim

> reviewed and verified locally.

Yet you opened [67 Pull requests](https://github.com/search?q=type%3Apr+author%3Aapoorvdarshan+created%3A2026-07-09++&type=pullrequests) with similar claims, on the same day as this one, in a 4-hour period.

Please don't include us in your LLM-spam list.

PR #877, opened 2026-07-13
Re-lease delayed messages before RabbitMQ's consumer_timeout

davidt99:
RabbitMQ requeues any delivery left unacked past its consumer_timeout (default 30 min, enforced since 3.8.15 on any queue type). Dramatiq holds delayed/retried messages unacked in memory until their eta, so a delay longer than the timeout is requeued by the server before it runs, churning the delay queue (#827).

Add an optional consumer_timeout to RabbitmqBroker. When set, it is sent as an x-consumer-timeout consumer argument on the delay-queue consumer only (honoured by RabbitMQ 3.12+), and the worker re-leases a held delayed message at 0.75x that value -- re-enqueuing it with its remaining delay so the eta is preserved and it still fires exactly once. Worker-side and queue-type agnostic; defaults to None (unchanged), minimum 15 minutes.

Addresses #827.

### DIFF
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -148,6 +148,41 @@ failover if the currently connected node fails.
 .. _connection parameters: https://pika.readthedocs.io/en/0.12.0/modules/parameters.html
 
 
+.. _consumer-timeouts:
+
+Consumer Timeouts and Long Delays
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RabbitMQ
+~~~~~~~~
+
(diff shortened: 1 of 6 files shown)


davidt99:
@LincolnPuzey as agreed, I'm only handling the consumer timeout. Although the behavior is relevant for rabbitmq, it exists on other potential brokers.
I think the free-thread build is flaky, the branch passed on my environment

davidt99:
@LincolnPuzey Any update? I'm aware of the consumer timeout in 4.3, but to be honest, there is no one solution, specifically when there is no stable API to get the rabbitmq server version with ampq 0.9.1.

PR #878, opened 2026-07-16
Asyncio: Prevent Exceptions raised from async actors from being caught in reference cycles

LincolnPuzey:
Asyncio middleware: Clear reference to `future` so that Exceptions are not caught in reference cycles

This should ensure that Exceptions reach zero references and are promptly deallocated,
rather than being caught in cycles, and relying on the garbage collector to destroy them.

This should improve memory usage in workers that run lots of async actors that raise Exceptions.

### DIFF
--- a/dramatiq/asyncio.py
+++ b/dramatiq/asyncio.py
@@ -179,3 +179,11 @@ class EventLoopThread(threading.Thread):
             if not done.wait(timeout=1.0):
                 raise RuntimeError("Timed out while waiting for coroutine.") from e
             raise
+
+        finally:
+            # Clear reference to the future, to prevent an Exception that was raised in the coroutine
+            # (and subsequently raised from a future.result() call)
+            # from being caught in a reference cycle.
+            del future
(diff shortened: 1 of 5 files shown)


LincolnPuzey:
Alternative fix to #806

LincolnPuzey:
@frankie567 I got the test reliably failing, then added a (IMO) simpler fix which passes it.

If you want to, please have a look and/or test it out to see if it improves your memory issue.

frankie567:
Sounds sensible, will try your branch in our environment, but that looks promising 🙂

Thank you for tackling this!

LincolnPuzey:
@frankie567 Did you get a chance to do any testing?

frankie567:
We run it for almost 3 weeks now in production, and the issue didn't come back. So yes, I think it's working great :) Thank you for all the work on this matter 🙏

LincolnPuzey:
@frankie567 Ok thanks, that's great to hear

PR #883, opened 2026-08-04
Share a single connection across all RabbitMQ consumers

### LINKED ISSUE #649: Dramatiq uses more RabbitMQ connections than necessary
We use RabbitMQ as our Dramatiq broker, and we're always dangerously close to the connection limit (which is 500 in our case). This is because Dramatiq creates a connection in each worker process for each queue we have, and we have a queue per Dramatiq task.

We think Dramatiq can be modified to use fewer connections. According to [CloudAMQP's RabbitMQ best practices](https://www.cloudamqp.com/blog/part1-rabbitmq-best-practice.html#connections-and-channels), you should generally limit connections to one connection per process, and use channels within that connection for per-thread use. Instead of doing this, Dramatiq creates a new connection with a single channel for each RabbitMQ consumer it creates, and it creates a consumer for each queue per process, when it could probably be made to just use one connection which each consumer within a process shares, and then each consumer can just create a channel for itself.

## What OS are you using?

Debian 12 (via the official Python Docker image)

v1.17.0

jholaj:
Fixes #649.

Every consumer opens its own connection (one per Q, plus one for its DQ), so a process consuming many queues holds a lot of mostly idle connections - a few hundred queues means several hundred connections per worker process.

With this change all consumers share one connection. pika isn't thread safe, so a small I/O thread owns it - deliveries go into per-consumer buffers, acks go back through `add_callback_threadsafe` (same as before). Each consumer still has its own channel, so prefetch works the same.

Only real difference: if the connection dies, all consumers of the process restart at once instead of one by one.

Existing RabbitMQ tests pass unchanged (cpython + gevent, RabbitMQ 4.2.9), and I added a test that consumers really do share one connection now.

### Note
While measuring this I found a second source of connections: `consume()` declares the queue through the calling thread's publisher connection, and since consumers are created from consumer threads, each of them also kept an
idle publisher connection around forever. `consume()` now drops that connection after the declare - `test_rabbitmq_broker_connections_are_lazy` is updated accordingly (it asserts even lazier behavior than before).

Happy to rework any part of this or just talk it through - I use dramatiq daily now and like the project a lot, so I'm glad to help however fits.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -80,3 +80,4 @@ of those changes to CLEARTYPE SRL.
 | [@ABolouk](https://github.com/ABolouk)                 | Amirhossein Bolouk Asli|
 | [@JoaoAlmeida20](https://github.com/JoaoAlmeida20)     | João Almeida           |
 | [@albcunha](https://github.com/albcunha)               | Alberto Cunha          |
+| [@jholaj](https://github.com/jholaj)                   | Jaroslav Holaj         |
--- a/dramatiq/brokers/rabbitmq.py
+++ b/dramatiq/brokers/rabbitmq.py
@@ -22,7 +22,8 @@ import os
 import time
 from functools import partial
 from itertools import chain
-from threading import Event, local
+from queue import Empty, Queue
+from threading import Event, Lock, Thread, local
 from typing import Any, Optional, Union
 
 import pika
@@ -152,10 +153,24 @@ class RabbitmqBroker(Broker):
         # Also, should only contain 'canonical' queue names.
         self.queues_pending: set[str] = set()
         self.state = local()
+        self._consumer_connection: Optional[_ConsumerSharedConnection] = None
+        self._consumer_connection_lock = Lock()
 
     @property
     def consumer_class(self):
         return _RabbitmqConsumer
+
+    def _get_consumer_connection(self):
+        """The connection shared by all of this broker's consumers.
+
+        Each consumer gets its own channel on this connection, so the
+        connection count per worker process stays constant no matter how
+        many queues are consumed.
+        """
+        with self._consumer_connection_lock:
+            if self._consumer_connection is None:
+                self._consumer_connection = _ConsumerSharedConnection(self)
+            return self._consumer_connection
 
     @property
     def connection(self):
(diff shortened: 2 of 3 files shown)


LincolnPuzey:
Hi, thanks for the PR. I haven't looked at the code yet.

I have seen pika has recently added [ThreadSafeConnection](https://github.com/pika/pika/blob/1d570cac7b925c77c87d1a20b0a24199539b7f45/pika/adapters/thread_safe_connection.py#L1296) which sounds like it works in a similar manner to your PR, in regards to running the connection in its own background thread.

Perhaps we should wait for that to be released, and update Dramatiq to use that, rather than implementing the same thing ourselves.

LincolnPuzey:
> consume() declares the queue through the calling thread's publisher connection, and since consumers are created from consumer threads, each of them also kept an
idle publisher connection around forever.

I think the `ConsumerThread` uses its publisher connection (on the `Broker`) to publish delayed messages. See [here](https://github.com/Bogdanp/dramatiq/blob/b0e96f9aa97c9bf4ab7101a6ce1d3a5205141c29/dramatiq/worker.py#L360). So I don't think that connection is idle forever.

jholaj:
Hello, sorry for the slow reply, I haven't had time lately.

Agreed - it's the same mechanism as this PR and better solved directly in pika. It won't cover everything though: reconnection and the consumer lifecycle stay in dramatiq either way, so a rewrite would drop the plumbing but keep most of the logic. It'd also mean bumping the pika floor from 1.0.

> I think the ConsumerThread uses its publisher connection (on the Broker) to publish delayed messages. See [here](https://github.com/Bogdanp/dramatiq/blob/b0e96f9aa97c9bf4ab7101a6ce1d3a5205141c29/dramatiq/worker.py#L360). So I don't think that connection is idle forever.

You're right that delay-queue consumer threads publish when `handle_delayed_messages` re-enqueues, so "idle forever" was an overstatement from me. The behavior is still correct though: the `connection` property is lazy, so the first re-enqueue simply reopens a connection, which is then kept and reused. Only the plain-queue consumer threads, which never publish, stop holding one forever. I'll fix the  comment. 👍

Fine by me to wait - if you agree, I'll update this PR once it's released. :)

